PR #4, opened 2018-07-31
Added support for custom engine.

Mulugruntz:


Mulugruntz:
@adamrehn Please note that for now, this has not been tested yet.

adamrehn:
Thanks for submitting this! Due to the refactoring that was involved in implementing the prerequisites for this functionality, the actual commit from this pull request can't be merged in, but I have implemented the same functionality in commit 7f682cd and tested that it functions correctly. The command-line syntax is slightly more verbose than what had been proposed, but has the benefit of being more explicit and also facilitating simpler validation logic.

Cheers,
Adam

PR #14, opened 2018-12-07
Update cuda docker image version

Clockwork-Muse:


adamrehn:
I've been contemplating adding CUDA 10 support for some time now, but I'm hesitant to force a newer version on users since this also forces the associated minimum driver version for CUDA 10 upon them, and has the potential to break existing code that uses the container images. As an alternative to changing the default CUDA base image, I've modified the behaviour of the `--cuda` flag in commit 0933fa0 to accept an optional version number that allows the user to select the CUDA version that they wish to use. This usage is depicted below:

```bash
# Uses the CUDA 9.2 base image
ue4-docker build RELEASE --cuda

# Uses the CUDA 10.0 base image
ue4-docker build RELEASE --cuda=10.0
```

The documentation will be updated to reflect this change once the next version of the ue4-docker package is published on PyPI, which I'll be doing just as soon as I've completed some further testing of the features that have been added since the previous release.

Clockwork-Muse:
Fair enough.

PR #39, opened 2019-06-04
Fix exclude templates

gcandal:


adamrehn:
Whoops, that's a silly typo on my behalf. Thanks for catching that!

adamrehn:
I just realised that I'd made the same mistake in the Windows version of that script too. Both versions are now fixed in ue4-docker 0.0.34.

Issue #22, opened 2019-03-26
GitHub credentials incorrect


herman-rogers:
So just parking fairly common issue, since we can only specify git credentials using our username and password this method will fail if you have 2-factor authentication setup. The simple work around is just to disable 2-factor in your account and the proper fix is to allow using ssh authentication (I don't think at the moment 2-factor for GitHub is enabled via cli but I might be wrong).

TBBle:
You can generate a [Personal Access Token](https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line) to use as a password in this case.

adamrehn:
Thanks @TBBle, it's good to know that the access token option is available for users who have 2FA enabled. I've added [a note about it](https://adamrehn.com/docs/ue4-docker/building-images/advanced-build-options#specifying-git-credentials) to the documentation to make it easier for people to find.

Support for SSH authentication is an interesting option, although it would require a bit of a re-think in terms of how the credential manager is implemented. I'll have a look into how feasible it would be to add an option for it and post a comment here with my findings.

herman-rogers:
Thanks for the tip @TBBle - I'll definitely go down that route in the meantime!

TBBle:
Would the credential manager rethink be easier if it was implemented as a [Credential Helper](https://git-scm.com/docs/api-credentials#_credential_helpers) rather than `SSH_ASKPASS` helper? I'm not totally clear if there's a difference in functionality there, as `SSH_ASKPASS` doesn't appear clearly-documented.

I've been thinking about this myself. My use-case is running on AWS, but if I was running this locally, then I'd like it to read from my SSH Agent in the SSH case.

And at some point, I suspect I'll need to be able to source from Perforce rather than git, which can work off user/pass and off tickets. Perforce is probably distinct-enough that it would need its own credential-managing approach anyway, so that's low-value here.

adamrehn:
@TBBle evidently forwarding of SSH credentials is now possible using experimental build features in the latest builds of Docker: <https://medium.com/@tonistiigi/build-secrets-and-ssh-forwarding-in-docker-18-09-ae8161d066>

Once BuildKit becomes the default build implementation for Docker and support is added for Windows containers, I'll revisit the credential management code in ue4-docker and overhaul it to use the new functionality, which should allow for a far wider variety of authentication mechanisms and source control systems.

adamrehn:
[...]

PR #47, opened 2019-08-16
Local ue4 source support and configurable tmp folder

jit-rc:
Unreal Engine can now be added from local copy, either a directory or Release archive file on the host machine using the argument  --ue-src 

tmp folder might require more space than might be available on normal OS tmp folder, to override that, now we can use --tmp-dir


adamrehn:
@jit-rc could you please describe the use case(s) that would benefit from working with source code from the host filesystem rather than a version control system?

The most obvious scenario I can envisage where code isn't checked into version control is when a developer is testing Engine modifications prior to committing them, but that use case would be far better served by performing incremental builds in an interactive container that bind-mounts the Unreal Engine source tree, rather than performing builds in a Dockerfile.

(I also realise that copying from the host filesystem could be used as a stop-gap solution for working with currently unsupported version control systems such as Perforce, but implementing proper support for those systems would be the correct solution in that case.)

jit-rc:
I was making the docker for https://github.com/carla-simulator/carla/tree/0.9.6/Util/Docker and don't need to change the UE code. The issue was github authentication would not work with 2-factor, and I was building the docker on a server with more storage than my laptop and didn't want to store and of my credentials there, it seemed simpler to share the downloaded release archive. Did it for my organizations use and shared it back if you feel that it could help others

adamrehn:
GitHub two-factor authentication actually [can be used with ue4-docker](https://adamrehn.com/docs/ue4-docker/building-images/advanced-build-options#specifying-git-credentials) if you generate a personal access token and specify the token in place of a password. (See issue #22.)

I'm hesitant to add yet another Dockerfile to ue4-docker in its current state, since the build infrastructure was only originally designed for a small number of images and has already gotten a little messy as I've split images and added new ones. That being said, I'd like to be able to accommodate a wider variety of configurations in the future, with the option for users to customise the process to a far greater extent than is currently possible. I've started a brainstorming discussion in issue #48 to start the ball rolling on a refactor of ue4-docker that would make the integration of functionality such as that presented in this pull request feasible in a clean way.

jit-rc:
Yes, I know about the personal access token, I didn't want to create and share that too on our shared server, though it is private server. I understand your hesitation and agree that there are too many docker files and redundancies  for windows and linux, let me just keep the changes on my local branch and you can close the PR without merging.

PR #68, opened 2020-02-11
Added missing libraries (enable CEF support)

galeone:
Adding these libraries (with the exception of `apt-file`, that I suggest keeping because it is really useful into the container) allow building applications that are using `libcef.so`

### DIFF
--- a/ue4docker/dockerfiles/ue4-full/linux/Dockerfile
+++ b/ue4docker/dockerfiles/ue4-full/linux/Dockerfile
@@ -61,5 +61,14 @@ RUN apt-get install -y --no-install-recommends \
 	x11-xkb-utils \
 	xauth \
 	xfonts-base \
-	xkb-data
+	xkb-data \
+	apt-file \
+	libasound2 \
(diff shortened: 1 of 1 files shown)


adamrehn:
There are a couple of minor issues with your library list:

- `libpango-1.0` is not actually a package, it gets treated as a regex and expanded automatically to the correct package name, which is `libpango-1.0-0`

- `libcairo-5c0` is not necessary, what you actually need is `libcairo2`, which it depends on

Although I agree that `apt-file` is useful, it's only really necessary when hunting down missing libraries, which only a small handful of users are likely to do. Rather than add it to the Dockerfile and increase the size for all users, I've added a note about using it to the [Writing Dockerfiles](https://unrealcontainers.com/docs/obtaining-images/write-your-own) page of the Unreal Containers community hub documentation.

I've added the required libraries in commit 3f5b10e, which is now live in ue4-docker version 0.0.51.

galeone:
Thank you very much!

PR #129, opened 2021-02-17
Upgrade pip in linux full image

floriankarydes:
`ue4-cli` fails to install inside image because it depends on `cryptography` and the later require `pip` to be upgraded.


adamrehn:
Thanks for your contribution!

PR #131, opened 2021-03-02
Disable GitDependencies.exe caching to .git/ue4-gitdeps

slonopotamus:
This commit reduces amount of data written to disk during ue4-source image build by ~9GB.


slonopotamus:
Note that I did *not* actually test this change within ue4-docker because I'm not sure how I run ue4-docker from source.

adamrehn:
Thanks for your contribution!

PR #135, opened 2021-03-04
Require semver 2.x.x

slonopotamus:
If there's no semver python module installed and user does `python setup.py develop --user` in ue4-docker repo in order to work on ue4-docker, then semver 3.x.x is pulled in.

1. Semver [website](https://github.com/python-semver/python-semver) clearly says that semver 3.x.x is not for production use until 3.0.0 is released
2. ue4-docker fails to import semver.py because there is no such file in semver 3.x.x

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ setup(
 		'packaging>=19.1',
 		'psutil',
 		'requests',
-		'semver>=2.7.9',
+		'semver>=2.7.9,<3.0.0',
 		'setuptools>=38.6.0',
 		'termcolor',
 		'twine>=1.11.0',


slonopotamus:
I'm not actually sure why `pip install ue4-docker` doesn't pull in semver-3, but anyway we don't want stuff to suddenly break when they release a 3.0.0 release.

adamrehn:
Thanks for your contribution!

slonopotamus:
I found out that `python setup.py develop --user` is not a good way to set up development from source. Instead, one should prefer to execute `pip install -e . --user` from repository root. The latter doesn't pick up prerelease dependencies (like `3.0.0.dev2`). Though the argument that "things will break the moment semver releases final 3.x.x release" is still valid, so this PR has a value.

PR #136, opened 2021-03-09
Fix dockerfiles to actually remove unneeded directories on Windows

slonopotamus:
It turns out that `del /s` is **not** deleting directory recursively

---

I noticed that UBT is still spending a lot of time checking for Git changes when building `ue4-minimal`. Inspection of container contents showed that `C:\UnrealEngine\.git` directory was still present in container even though it was supposed to be removed by the time engine is built.

Manual testing inside container showed that `del /s /q C:\UnrealEngine\.git` actually does *not* delete neither directory nor Git files inside it.

This PR switches to `rmdir /s /q` that **does** fully delete directory and its contents recursively.

### DIFF
--- a/ue4docker/dockerfiles/ue4-minimal/windows/Dockerfile
+++ b/ue4docker/dockerfiles/ue4-minimal/windows/Dockerfile
@@ -9,7 +9,7 @@ FROM ${NAMESPACE}/ue4-source:${TAG}-${PREREQS_TAG} AS builder
 {% endif %}
 
 # Remove the .git directory to disable UBT `git status` calls and speed up the build process
-RUN del /s /q C:\UnrealEngine\.git
+RUN rmdir /s /q C:\UnrealEngine\.git
 
 # Set the changelist number in Build.version to ensure our Build ID is generated correctly
 COPY set-changelist.py C:\set-changelist.py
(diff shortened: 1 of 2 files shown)


adamrehn:
Thanks for your contribution!

slonopotamus:
Ouch. This change broke building engine with `--exclude ddc`:

```
BUILD SUCCESSFUL
AutomationTool exiting with ExitCode=0 (Success)
The command 'cmd /S /C .\Engine\Build\BatchFiles\RunUAT.bat BuildGraph -target="Make Installed Build Win64" -script=Engine/Build/InstalledEngineBuild.xml -set:HostPlatformOnly=true -set:WithDDC=%BUILD_DDC%  &&       rmdir /s /q C:\UnrealEngine\LocalBuilds\InstalledDDC 2>NUL && echo. && echo.RUN directive complete. Docker will now commit the filesystem layer to disk. && echo.Note that for large filesystem layers this can take quite some time. && echo.Performing filesystem layer commit... && echo.' returned a non-zero code: 2
[ue4-docker build] Error: failed to build image "adamrehn/ue4-minimal:4.26.1-ltsc2019".
```

It seems like `rmdir` returns exitcode=2 if dir doesn't exist.

slonopotamus:
Seems like `rmdir /s /q C:\UnrealEngine\LocalBuilds\InstalledDDC 2>NUL` also needs `|| exit 0` trick. Though I'm not sure how to properly put this within a chain of commands combined with `&&`.

slonopotamus:
Created #139 with a fix for `--exclude ddc` issue.

PR #139, opened 2021-03-17
Fix engine build with --exclude ddc failing to remove nonexistent directory

slonopotamus:
It turns out that rmdir returns exitcode=2 if directory to delete doesn't exist.
So, only delete DDC directory if it actually exists.

Broken by 9cf1375c33098a2787b5f514fd0ff36167a12b96.

----

Is this directory actually worth deleting? There are tons of other intermediate build products inside the image that are still left behind. And this intermediate image is going to be thrown away completely a couple of steps later when ue4-minimal image is produced.


slonopotamus:
OK, I've just confirmed that this does fix installed engine build.

adamrehn:
Thanks for identifying this issue and contributing a fix! For memory, that directory is one of the largest ones in the intermediate image, so the rationale for removing it is that doing so reduces the amount of disk space required during the build process. Given that building the Windows container images requires hundreds of gigabytes of free disk space, my thinking is that any reduction we can make is certainly worth it.

PR #140, opened 2021-03-18
Drop pdbcopy workaround

slonopotamus:
It was needed for UE older than 4.20 because of https://issues.unrealengine.com/issue/UE-51362
Since 4.20 is the oldest supported version by ue4-docker now, this workaround is no longer needed

----

Not tested yet, my build with this change is in progress. Will comment if/when it finishes.


adamrehn:
Let me know once your tests have completed successfully and I'll be happy to merge this!

slonopotamus:
Okay, it turns out that [Visual Studio doesn't install pdbcopy at all](https://developercommunity.visualstudio.com/t/pdbcopyexe-missing-from-155/168414#T-N223034). We still could simplify the procedure a bit by just [installing debugging tools](https://github.com/adamrehn/ue4-docker/blob/v0.0.81/ue4docker/dockerfiles/ue4-build-prerequisites/windows/Dockerfile#L43) straight into ue4-build-prerequisites image and drop code that copies it over to various locations, but possibly it isn't worth it. Closing this PR.

Issue #141, opened 2021-03-19
Packaging project fails if missing write permission to /home/ue4/UnrealEngine


RoryDungan:
Output of the `ue4-docker info` command:

```
ue4-docker info
ue4-docker version:         0.0.79 (latest available version is 0.0.81)
Operating system:           Linux (Fedora 30 (Workstation Edition), 5.3.7-200.fc30.x86_64)
Docker daemon version:      19.03.12
NVIDIA Docker supported:    No
Maximum image size:         No limit detected
Available disk space:       1.52 TiB
Total system memory:        31.17 GiB physical, 0 bytes virtual
Number of processors:       4 physical, 8 logical
```

Additional details:

- Are you accessing the network through a proxy server? **No**

-----------------------------------

Hi! I'm using ue4-docker for containerised builds on Jenkins. The Jenkins Docker plugin always runs containers with this command: 
```
docker run -t -d -u 570896598:570800513 -w /var/jenkins/workspace/dev_game_test -v /var/jenkins/workspace/dev_game_test:/var/jenkins/workspace/dev_game_test:rw,z -v /var/jenkins/workspace/dev_game_test@tmp:/var/jenkins/workspace/dev_game_test@tmp:rw,z
```
The arguments after `-u` mach the _host_ user and group ID, and `dev_game_test` is replaced with the Jenkns job name. The user seems to be set to that in order to ensure that the container has write permission to the mounted volume so that it can save build artifacts there. However, this means that since we aren't running as the `ue4` user any more we lose write permission to `/home/ue4/UnrealEngine`. 

This shouldn't cause any issues, since it's normal to run programs from a read-only directory. However, the way that the ue4-docker images are set up, UE4 will actually set up a bunch of symbolic links to Mono DLLs inside `/home/ue4/UnrealEngine/Engine/Binaries/ThirdParty/Mono/Linux`, as well as compile some DLLs inside `/home/ue4/UnrealEngine/Engine/Intermediate/Build/BuildRules` the first time it needs to run the UnrealAutomationTool. 

[...]

Issue #138, opened 2021-03-16
Dynamically load list of Windows base images


luc122c:
I'm currently running an insider build of Windows so I can't seem to build Windows containers as the latest available tag is 20H2. The list of valid tags seems to be [defined here](https://github.com/adamrehn/ue4-docker/blob/7a2fb52df9b14ddb292866f7b09ed76927a8ef63/ue4docker/infrastructure/WindowsUtils.py#L18). My OS version is 2009 which seems to be mapped to 20H2 [here](https://github.com/adamrehn/ue4-docker/blob/7a2fb52df9b14ddb292866f7b09ed76927a8ef63/ue4docker/infrastructure/WindowsUtils.py#L161), however when I try to build, I get the following errors:

```
[ue4-docker build] Error: building images with a different kernel version than the host, 
[ue4-docker build] but a custom DLL directory has not specified via the `-dlldir=DIR` arg.
[ue4-docker build] The DLL files will be the incorrect version and the container OS will 
[ue4-docker build] refuse to load them, preventing the built Engine from running correctly.
```

Microsoft publish a list of their tags in JSON format so it should be fairly straight forward to pull this in and parse it i'd imagine. I'd make a PR if I knew Python! The list can be found here: [https://mcr.microsoft.com/v2/windows/tags/list](https://mcr.microsoft.com/v2/windows/tags/list). It does include tags for 2009:

- 2009
- 2009-KB4579311
- 2009-KB4579311-amd64
- 2009-amd64

Output of the `ue4-docker info` command:

```
$ ue4-docker info
ue4-docker version:         0.0.81 (latest available version is 0.0.81)
Operating system:           Windows 10 Pro for Workstations Version 2009 (OS Build 21332.1010)
Docker daemon version:      20.10.5
NVIDIA Docker supported:    No
Maximum image size:         20GB
Available disk space:       18.74 GiB
Total system memory:        31.92 GiB physical, 36.67 GiB virtual
Number of processors:       4 physical, 8 logical
```

slonopotamus:
Okay, there is a dynamic list of Windows tags. But how ue4-docker understands which windowsservercore version matches?

luc122c:
This list is from that page. Presumably for testing the user could specify with the base tag flag?
`-basetag BASETAG      Windows Server Core base image tag to use for Windows
                        containers (default is the host OS version)`

TBBle:
[...]

PR #144, opened 2021-03-26
Reduce ue4-build-prerequisites Windows image

slonopotamus:
This commit:

1. Replaces base image from mcr.microsoft.com/dotnet/framework/sdk to mcr.microsoft.com/windows/servercore
The former is 9GB while the latter is just 5GB.
dotnet/framework/sdk contains lots of things that are not actually needed to build ue4,
the hugest is VS Test Agent + VS Build Tools 2019 that take together 2.2GB

2. Stops installing unneeded VS Build Tools workloads/components and only installs required minimum:
* MSBuild (needed to build MSBuild-based UE4 tools, like UBT)
* NuGet (required to run MSBuild and also [checked in GenerateProjectFiles.bat](https://github.com/EpicGames/UnrealEngine/blob/4.26.1-release/Engine/Build/BatchFiles/GenerateProjectFiles.bat#L31-L54))
* C# compiler (needed to build C# UE4 tools)
* .NET SDK 4.5 + 4.6.2 (same as above)
* Windows 10 SDK (needed to build C++ code)
* VCTools: C++ compiler, VC redist (same as above)

Cumulatively, this commit reduces ue4-build-prerequisites image from 24.5GB down to 10.7GB.
The most important consequence of this change is that it reduces ue4-minimal/ue4-full image sizes by the same delta

----

See https://github.com/adamrehn/ue4-docker/pull/144#issuecomment-810395548 for test matrix of this PR.

### DIFF
--- a/ue4docker/dockerfiles/ue4-build-prerequisites/windows/install-prerequisites.bat
+++ b/ue4docker/dockerfiles/ue4-build-prerequisites/windows/install-prerequisites.bat
@@ -10,7 +10,7 @@ call refreshenv
 @rem Forcibly disable the git credential manager
 git config --system credential.helper "" || goto :error
 
-@rem Install the Visual Studio 2017 Build Tools workloads and components we need, excluding components with known issues in containers
+@rem Install the Visual Studio 2017 Build Tools workloads and components we need
 @rem (Note that we use the Visual Studio 2019 installer here because the old installer now breaks, but explicitly install the VS2017 Build Tools)
 curl --progress -L "https://aka.ms/vs/16/release/vs_buildtools.exe" --output %TEMP%\vs_buildtools.exe || goto :error
 %TEMP%\vs_buildtools.exe --quiet --wait --norestart --nocache ^
@@ -19,12 +19,14 @@ curl --progress -L "https://aka.ms/vs/16/release/vs_buildtools.exe" --output %TE
 	--installChannelUri "https://aka.ms/vs/15/release/channel" ^
 	--channelId VisualStudio.15.Release ^
 	--productId Microsoft.VisualStudio.Product.BuildTools ^
-	--add Microsoft.VisualStudio.Workload.VCTools;includeRecommended ^
-	--add Microsoft.VisualStudio.Workload.ManagedDesktopBuildTools;includeRecommended ^
-	--add Microsoft.VisualStudio.Workload.UniversalBuildTools ^
-	--add Microsoft.VisualStudio.Workload.NetCoreBuildTools ^
+	--add Microsoft.VisualStudio.Workload.VCTools ^
 	--add Microsoft.VisualStudio.Workload.MSBuildTools ^
-	--add Microsoft.VisualStudio.Component.NuGet
+	--add Microsoft.VisualStudio.Component.NuGet ^
+	--add Microsoft.VisualStudio.Component.Roslyn.Compiler ^
+	--add Microsoft.VisualStudio.Component.Windows10SDK.17763 ^
+	--add Microsoft.Net.Component.4.5.TargetingPack ^
+	--add Microsoft.Net.Component.4.6.2.SDK ^
+	--add Microsoft.Net.Component.4.6.2.TargetingPack
 python C:\buildtools-exitcode.py %ERRORLEVEL% || goto :error
 
 @rem Copy pdbcopy.exe to the expected location(s)
--- a/ue4docker/infrastructure/BuildConfiguration.py
+++ b/ue4docker/infrastructure/BuildConfiguration.py
@@ -231,7 +231,7 @@ class BuildConfiguration(object):
 		
 		# Store the tag for the base Windows Server Core image
 		self.basetag = self.args.basetag if self.args.basetag is not None else self.hostBasetag
-		self.baseImage = 'mcr.microsoft.com/dotnet/framework/sdk:4.8-windowsservercore-' + self.basetag
+		self.baseImage = 'mcr.microsoft.com/windows/servercore:' + self.basetag
 		self.prereqsTag = self.basetag
 		
 		# Verify that any user-specified base tag is valid


TBBle:
I like this approach. It has the extra benefit that it's feasible to build a ue4-prerequisites against Windows Insider Server Core base images, which don't exist for the mcr.microsoft.com/dotnet/framework/sdk, i.e. #138 would be resolvable.

I'd like to see *someone* (i.e. I'm not trying to push this all on you, @slonopotamus) test this for UE 4.20 onwards (i.e. the ue4-docker supported versions). I expect LTSC2019 and 20H2 is sufficient coverage; if they work, then anything between should work too, or can be dealt with when identified.

I'd also like to get some non-trivial projects built using the resulting images, e.g. [the UE4 Shooter Game](https://docs.unrealengine.com/en-US/Resources/SampleGames/ShooterGame/index.html), to make sure we're not breaking something that is always going to be needed later in the game's build-process than ue4-minimal.

To be fair, I'm not sure that mooted 'later' issue exists, but #141 was a very similar issue, where part of the UE4 engine build was not happening until the first time you tried to build a game. That was observed and fixed on Linux, but it's possible the same thing will happen on Windows, and if it (for some reason) turns out to to depend on, e.g., .NET Core SDK (which this change excludes) then we'd want to know that and fix it.

slonopotamus:
Additional thought: this PR allows us to control which exactly Windows SDK UE4 will be built against (because we install only a single Windows SDK now). We possibly want to change it to `Microsoft.VisualStudio.Component.Windows81SDK`.

TBBle:
Why Windows 8.1 SDK? The current UE4 docs [specify Windows SDK 10.0.18362 or newer](https://docs.unrealengine.com/en-US/ProductionPipelines/DevelopmentSetup/VisualStudioSetup/index.html).

UE 4.20.0 [preferred 10.0.16299.0](https://github.com/EpicGames/UnrealEngine/blob/4.20.0-release/Engine/Source/Programs/UnrealBuildTool/Platform/Windows/UEBuildWindows.cs#L319) and in UE 4.25.0 that was changed to [prefer 10.0.18362.0, then 10.0.16299.0](https://github.com/EpicGames/UnrealEngine/blob/4.25.0-release/Engine/Source/Programs/UnrealBuildTool/Platform/Windows/UEBuildWindows.cs#L627-L631).

I also noticed that 4.25.0 is also when they started preferring the VS 2019 toolchain (14.2X), since that's defined in the same file, but they still list VS 2017 toolchains (14.1X) as 'preferred' there.

So we _probably_ should be using the 10.0.16299.0 Windows SDK by default, since that's a preferred version for all ue4-docker-supported UE4 versions.

slonopotamus:
> Why Windows 8.1 SDK?

First, because this is what gets installed if one chooses 'Unreal Engine' component in Visual Studio Installer.

Second, we need to talk about SDK compatibility. As far as I know, if you build your app against SDK version N, all its functionality is guaranteed to be present on Windows builds N and later. But the opposite is not true. If you use some function present in version N, you may discover that it doesn't exist at all on older Windows builds. So, the safest strategy is to build against the oldest supported SDK (and Windows 8.1 is not dead yet).

On the other side, it looks like UE4 *itself* doesn't use any specific functions from Windows 10 SDK and happily builds with any of them. So the only possible source of incompatibilities is game code.

Now, why I specified 17763 in this PR: 17763 is the version that was effectively installed by previous build-prerequisites code.

So, I see two strategies here:
1. Use the oldest (8.1). If app compiles, it runs, but game cannot use stuff from newer SDKs.
2. Use the latest (17763). All modern SDK stuff is available for game, but it is game responsibility to guard with version checks in runtime because underlying OS might not have support.
3. Use what Epic claim to be preferred. But since we save a single ue4-build-prerequisites for all engine versions currently, it isn't clear what engine version to look at. I also wonder why UE4 prefers some SDK versions over others.

slonopotamus:
WRT Windows SDK 10.0.18362: it looks like that [it isn't available in VS Build Tools 2017](https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2017#visual-c-build-tools). 17763 is the last one that is listed.

slonopotamus:
UPD: I've also successfully built UE4 **4.20.3** on both 2019ltsc and 20H2 using changes from this PR.

TBBle:
I'm pretty sure the 'preferred' versions are what Epic are using on their CI system, so that'll match both what they QA, and what is in the Epic-distributed UE4 Installed Build version. Trying to match compilers for a project that works with both ue4-docker and the Epic-distributed build is how I came across this in the first place. I believe they mentioned it in their release notes at one point.

slonopotamus:
So, do you want me to switch to 16299? Or we just leave 17763 (at least for this PR) because it matches how ue4-docker worked before?

TBBle:
I'd leave it for now. I suspect 17763 is the default for the MS build tools installer because that's what Windows Server LTSC 2019 is using, which is still the most-recent Windows Server LTSC release.

Personally, I lean towards using a new Windows SDK, because that enables functionality that you can't enable otherwise, while still being able to choose to not use functionality that will not work on older versions of Windows. If you use an older SDK to guarantee compatibility with older (upstream-unsupported) versions of Windows, then you cannot opt-in to newer functionality.

I recognise this is not an uncontroversial position, so I think the easiest thing is to leave it as-is until more site-specific tuneable elements are exposed, and we move to the VS 2019 Build Tools installer channel, which (from memory) will allow finer-grained installation choices too, due to an improved dependency tree.

slonopotamus:
Okay, then I'm leaving it as-is, with 17763.

adamrehn:
@slonopotamus for a big change to the build prerequisites image like this, we'll need full test coverage before I can feel confident merging the pull request. That includes building all supported versions of the Unreal Engine (4.20 through 4.26) on both the latest LTSC and SAC releases of Windows, and verifying that all produced images are able to successfully package Unreal projects. If you're building the `ue4-full` image then the [ue4-docker test](https://docs.adamrehn.com/ue4-docker/commands/test) command can be used to run these tests.

Given that this pull request also changes the base image to plain Windows Server Core rather than any .NET Framework runtime image, I'd also be interested in knowing exactly which versions of Windows are and are not affected by the .NET issue discussed on Microsoft's [Known issues for containers](https://docs.microsoft.com/en-us/visualstudio/install/build-tools-container-issues?view=vs-2019#windows-container) page:

> If you base your image directly on microsoft/windowsservercore, the .NET Framework might not install properly and no install error is indicated. Managed code might not run after the install is complete. Instead, base your image on microsoft/dotnet-framework:4.7.1 or later. As an example, you might see an error when building with MSBuild that's similar to the following:
> 
> C:\BuildTools\MSBuild\15.0\bin\Roslyn\Microsoft.CSharp.Core.targets(84,5): error MSB6003: The specified task executable "csc.exe" could not be run. Could not load file or assembly 'System.IO.FileSystem, Version=4.0.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.

The fact that Microsoft's documentation still includes the details of that issue suggests that one or more currently-supported versions of Windows still suffer from it. If that turns out to just be Windows Server 2016 then we can safely ignore it, since that particular LTSC release has a variety of known bugs and I'm happy to drop support for it anyway as discussed in <https://github.com/adamrehn/ue4-docker/issues/99#issuecomment-791190687>, but if there are newer SAC releases that are affected too then we need to know.

slonopotamus:
> If you base your image directly on microsoft/windowsservercore, the .NET Framework might not install properly and no install error is indicated. 

I have no idea what host/guest/Docker versions this is talking about. It could possibly be fixed in some updates so we will never hit it. I am running my tests on fully-updated systems.

> If you're building the ue4-full image then the `ue4-docker test` command can be used to run these tests.

Okay, I can do that.

> That includes building all supported versions of the Unreal Engine (4.20 through 4.26) on both the latest LTSC and SAC releases of Windows

I am afraid I don't have enough hardware to do that in any adequate amounts of time. I have only two boxes here, one builds `ue4-minimal` in ~17h while the other takes ~21h. What's worse, the faster of them **has** to stay on Windows 20H2, so I'd need to juggle with reinstalling different Windows versions on the slower one in order to test through them. So, building 7 engine versions across 7 Windows versions is going to take *some* time with my resources.

TBBle:
https://github.com/MicrosoftDocs/visualstudio-docs/pull/722 is the discussion about servercore-based issue, it specifically was known to affect the VS 2017 15.6 installer. There was supposed to be a follow-up to the note when the bug was fixed, but that seems to have been forgotten.

It does seem, from that discussion, to relate to the installer, not the base image, so if it turns out it wasn't fixed until the VS2019 installer, we would need to switch over to that first, which shouldn't be too painful.

slonopotamus:
> we would need to switch over to that first, which shouldn't be too painful.

We already use VS2019 installer since 2ff06d1063f7010da37b982d011c247bf1f5a93c

TBBle:
Ah, sorry, you're right. I confused myself, thinking of the VS 2017 _channel_.

TBBle:
I propose that 2004 isn't a strongly-interesting test, because it's basically the same kernel as 20H2. So maybe now's the time to properly discuss dropping LTSC2016 (https://github.com/adamrehn/ue4-docker/issues/99#issuecomment-791190687), even though it's still supported upstream until January 2022.

slonopotamus:
Okay, nice, I hit #146 on the first run.

slonopotamus:
I've tested 4.20 on 20H2. It half-passes `ue4-docker test`: `build-and-package.py` passes while `consume-external-deps.py` fails because of https://github.com/adamrehn/conan-ue4cli/issues/12. I'm going to count this as "passes".

slonopotamus:
4.26 passes `ue4-docker test` on both 20H2 and 2019 LTSC if #149 is applied.

slonopotamus:
Note to myself: for unknown reasons, 4.23 built on 20H2 in just 12h instead of 17h for 4.25/4.26. This is worth investigating, possibly newer engine versions could be fixed to build as fast?

slonopotamus:
Lo and behold, 20H2 testing is finished. Results are: 4.20 suffers from https://github.com/adamrehn/conan-ue4cli/issues/12. 4.25 and 4.26 suffer from https://github.com/adamrehn/ue4-docker/issues/146. Both problems are unrelated to this PR.

slonopotamus:
> So maybe now's the time to properly discuss dropping LTSC2016

@TBBle, here you are: #151

slonopotamus:
Given current [results](https://github.com/adamrehn/ue4-docker/pull/144#issuecomment-810395548), I'm not expecting anything new to break. Maybe it's time to merge this PR? @adamrehn. Anyway, I'm continuing builds.

slonopotamus:
[Testing](https://github.com/adamrehn/ue4-docker/pull/144#issuecomment-810395548) is finished!

PR #145, opened 2021-03-30
Mark as EOL Windows releases no longer supported by Microsoft

slonopotamus:
According to [Base image servicing lifecycles](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle), these Windows Server Core releases are already EOL as of today


slonopotamus:
Whoops, I didn't mean to close PR.

PR #147, opened 2021-03-31
Add diagnostic for 20GiB COPY Docker bug on Windows

slonopotamus:


ue4docker/dockerfiles/diagnostics/20gig/windows/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# Create a file slightly over 20GiB
+RUN fsutil file createnew C:\bigfile.txt 22548578304
+
+FROM mcr.microsoft.com/windows/servercore:${BASETAG}
+
+# Add a sentinel label so we can easily identify intermediate images
+LABEL com.adamrehn.ue4-docker.sentinel="1"
+
+# Copy more than 20GiB between docker layers
+COPY --from=intermediate C:\bigfile.txt C:\\bigfile.txt
```
TBBle, line 17: Different backslashes on the two filenames. Is that meaningful?

ue4docker/dockerfiles/diagnostics/20gig/windows/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# escape=`
+ARG BASETAG
+FROM mcr.microsoft.com/windows/servercore:${BASETAG} as intermediate
+
+# Add a sentinel label so we can easily identify intermediate images
+LABEL com.adamrehn.ue4-docker.sentinel="1"
+
+# Create a file slightly over 20GiB
+RUN fsutil file createnew C:\bigfile.txt 22548578304
```
TBBle, line 9: This step will fail if the 8gB bug is present. Should we try to avoid that? Is it possible/reasonable to use one diagnostic to test for both issues, and distinguish the result?

slonopotamus: Oh, that's why you were using 3x7GiB. I dunno whether we want to distinguish them. What we actually want in the end is to be able to copy > 20GiB no matter how it is organized in files. I can easily switch to 3x7GiB is you think this is preferred.

ue4docker/dockerfiles/diagnostics/20gig/windows/Dockerfile
slonopotamus: > Is that meaningful?

Nope, fixed

ue4docker/dockerfiles/diagnostics/20gig/windows/Dockerfile
slonopotamus: Okay, I switched it to 3x7GiB.

TBBle (approved):
Looks good to me. Have you had a chance to confirm both failure with 20.03.x and success with a 21.xx (master) build of Docker?

slonopotamus:
I have Docker desktop 3.2.2 that includes 20.10.5 Docker Engine and it fails to pass this diagnostic. I haven't tried yet to build Docker from master (and not sure how I install it afterwards).

adamrehn:
@slonopotamus thanks for implementing this!

Issue #151, opened 2021-04-06
Status of Windows 2016 LTSC support


slonopotamus:
1. [Docker Desktop dropped support in 2.1.x](https://docs.docker.com/docker-for-windows/release-notes/#docker-community-edition-2002-2019-01-16):
> Windows 14393 is marked as deprecated ; it will not be supported anymore in the next major stable release (2.1.0.0 and further)

Latest Docker Desktop version that can be installed on 2016 LTSC was 2.0.0.3. That old version neither has fixes for 8GB nor for 20GB issues.

2. Microsoft mainstream support [ends on 2021-10-12](https://docs.microsoft.com/en-us/windows/release-health/release-information)

3. @adamrehn says he [stopped running tests on Windows 2016 LTSC some time ago](https://github.com/adamrehn/ue4-docker/issues/99#issuecomment-791190687).

I see two options here:
a. Drop support for Windows 2016 LTSC
b. Document how to install Docker on it and then do option A somewhere around October :)

slonopotamus:
UPD: I suddenly noticed that I am mixing Windows *Server* 2016 LTSC and non-server version.

The non-server has additional issue is that it lacks process isolation (or, at least, [ue4-docker documentation says so](https://docs.adamrehn.com/ue4-docker/configuration/configuring-windows-10):

> Starting in Windows 10 version 1809 it is possible to enable process isolation mode, but this must be done manually.

TBBle:
To be clear, we are meant to be talking about Windows Server LTSC 2016, right? The link in
>2. Microsoft mainstream support [ends on 2021-10-12](https://docs.microsoft.com/en-us/windows/release-health/release-information)

points at the Windows 10 lifecycles, which are different from the Windows Server lifecycles.

Windows Server LTSC 2016's "Mainstream End Date" is [11th January 2022](https://docs.microsoft.com/en-us/lifecycle/products/windows-server-2016) with "Extended End Date" ([security updates](https://docs.microsoft.com/en-us/lifecycle/policies/fixed#extended-support)) 5 years after that.

I don't think Docker Desktop is supported _at all_ on Windows Server, but I'm not certain of that claim. It probably works, but the support level will probably be "If it breaks, you get to keep both pieces". And the current version certainly won't run on Windows Server LTSC 2016, but can probably be beaten into working on Windows Server LTSC 2019 (Desktop Experience) if someone is so-motivated.

[...]

PR #153, opened 2021-04-07
Fix curl --progress flag

slonopotamus:
This commit fixes "curl: option --progress: is ambiguous" error

curl never had --progress flag. Instead, it has --progress-bar.
--progress worked before due to heuristic logic,
but not it became ambiguous.

See https://github.com/adamrehn/ue4-docker/issues/151#issuecomment-815141996 and https://github.com/moby/moby/pull/40914


TBBle (approved):

adamrehn:
@slonopotamus thanks for this!

Issue #133, opened 2021-03-03
ue4-build-prerequisites size bloat caused by multiple VS Build Tools


slonopotamus:
Currently, ue4-build-prerequisites [uses](https://github.com/adamrehn/ue4-docker/blob/v0.0.79/ue4docker/infrastructure/BuildConfiguration.py#L222) mcr.microsoft.com/dotnet/framework/sdk as base image. If we look into its Dockerfile, we can see that it already [contains](https://github.com/microsoft/dotnet-framework-docker/blob/main/src/sdk/4.8/windowsservercore-20H2/Dockerfile#L20-L30) VS **2019** Build Tools. On top of that, ue4-build-prerequisites [installs](https://github.com/adamrehn/ue4-docker/blob/v0.0.79/ue4docker/dockerfiles/ue4-build-prerequisites/windows/install-prerequisites.bat#L13-L28) VS **2017** Build Tools.

I either suggest:
a. Changing base image to something more lightweight (mcr.microsoft.com/dotnet/framework/sdk is almost 9GB)
b. Stop installing VS 2017 Build Tools toolchain completely and instead use VS 2019 Build Tools to build UE4.

slonopotamus:
Okay, the problem is that VS 2019 support was only added in UE-4.22 and ue4-docker claims to support UE >= 4.19...

slonopotamus:
Actually, option A is more interesting than I initially thought. This is described in https://adamrehn.com/articles/building-docker-images-for-unreal-engine-4/#issues-specific-to-windows-containers:

> Windows Server Core does not include the DirectX or OpenGL runtime files. Although the Windows 10 SDK that is installed by the Visual Studio 2017 Build Tools installer includes the DirectX SDK, it does not include all of the runtime DLL files that are required to run applications which have been built against the SDK. Both these DLLs and a set of OpenGL runtime files would ordinarily be provided by Windows, but this is not the case for Windows Server Core.

So, possibly switch to `mcr.microsoft.com/windows` image as a base would improve overall usability of ue4-docker.

slonopotamus:
@adamrehn So, I suggest first discussing which path you would like to go even if you don't currently have time to work on this. Having a decision would open an opportunity for a possible contributor to pick up this issue and work on it.

adamrehn:
> a. Change base image to something more lightweight (mcr.microsoft.com/dotnet/framework/sdk is almost 9GB) that doesn't include VS 2019 Build Tools

[...]

PR #173, opened 2021-05-27
Add support for Visual Studio 2019

### LINKED ISSUE #172: Add support for VS2019
Subj. This is a prerequisite for #171.

Note that I am already working on this feature.

slonopotamus:
resolves #172

----

This is not tested yet, I'm currently running builds.

The only questionable thing in this PR is a change to `ue4-build-prerequisites` tag. It effectively makes all `ue4-build-prerequisites` images currently uploaded to DockerHub unusable. We can avoid that by only appending `-vs<number>` to tag for VS2019, but I'm not sure this is a good approach.

### DIFF
--- a/ue4docker/dockerfiles/ue4-build-prerequisites/windows/install-prerequisites.bat
+++ b/ue4docker/dockerfiles/ue4-build-prerequisites/windows/install-prerequisites.bat
@@ -10,24 +10,28 @@ call refreshenv
 @rem Forcibly disable the git credential manager
 git config --system credential.helper "" || goto :error
 
-@rem Install the Visual Studio 2017 Build Tools workloads and components we need
-@rem (Note that we use the Visual Studio 2019 installer here because the old installer now breaks, but explicitly install the VS2017 Build Tools)
+set VISUAL_STUDIO_BUILD_NUMBER=%~1
+
+@rem Install the Visual Studio Build Tools workloads and components we need
+@rem NOTE: We use the Visual Studio 2019 installer even for Visual Studio 2017 here because the old installer now breaks
+@rem NOTE: VS2019 Build Tools doesn't have 4.6.2 .NET SDK and what actually gets installed is 4.8
 curl --progress-bar -L "https://aka.ms/vs/16/release/vs_buildtools.exe" --output %TEMP%\vs_buildtools.exe || goto :error
 %TEMP%\vs_buildtools.exe --quiet --wait --norestart --nocache ^
 	--installPath C:\BuildTools ^
-	--channelUri "https://aka.ms/vs/15/release/channel" ^
-	--installChannelUri "https://aka.ms/vs/15/release/channel" ^
-	--channelId VisualStudio.15.Release ^
+	--channelUri "https://aka.ms/vs/%VISUAL_STUDIO_BUILD_NUMBER%/release/channel" ^
+	--installChannelUri "https://aka.ms/vs/%VISUAL_STUDIO_BUILD_NUMBER%/release/channel" ^
+	--channelId VisualStudio.%VISUAL_STUDIO_BUILD_NUMBER%.Release ^
 	--productId Microsoft.VisualStudio.Product.BuildTools ^
 	--locale en-US ^
 	--add Microsoft.VisualStudio.Workload.VCTools ^
 	--add Microsoft.VisualStudio.Workload.MSBuildTools ^
 	--add Microsoft.VisualStudio.Component.NuGet ^
 	--add Microsoft.VisualStudio.Component.Roslyn.Compiler ^
+	--add Microsoft.VisualStudio.Component.VC.Tools.x86.x64 ^
 	--add Microsoft.VisualStudio.Component.Windows10SDK.17763 ^
 	--add Microsoft.Net.Component.4.5.TargetingPack ^
-	--add Microsoft.Net.Component.4.6.2.SDK ^
-	--add Microsoft.Net.Component.4.6.2.TargetingPack
+	--add Microsoft.Net.ComponentGroup.4.6.2.DeveloperTools
+
 python C:\buildtools-exitcode.py %ERRORLEVEL% || goto :error
 
 @rem Copy pdbcopy.exe to the expected location(s)
--- a/ue4docker/infrastructure/BuildConfiguration.py
+++ b/ue4docker/infrastructure/BuildConfiguration.py
@@ -27,6 +27,23 @@ DEFAULT_CUDA_VERSION = '9.2'
 # The default memory limit (in GB) under Windows
 DEFAULT_MEMORY_LIMIT = 10.0
 
+class VisualStudio(object):
+	VS2017 = '2017'
+	VS2019 = '2019'
+
+	BuildNumbers = {
+		VS2017 : '15',
+		VS2019 : '16',
+	}
+
+	MinSupportedUnreal = {
+		# Unreal Engine 4.23.1 is the first that successfully builds with Visual Studio v16.3
+		# See https://github.com/EpicGames/UnrealEngine/commit/2510d4fd07a35ba5bff6ac2c7becaa6e8b7f11fa
+		#
+		# Unreal Engine 4.25 is the first that works with .NET SDK 4.7+
+		# See https://github.com/EpicGames/UnrealEngine/commit/5256eedbdef30212ab69fdf4c09e898098959683
+		VS2019 : semver.VersionInfo(4, 25)
+	}
 
 class ExcludedComponent(object):
 	'''
@@ -77,6 +94,7 @@ class BuildConfiguration(object):
 		parser.add_argument('--exclude', action='append', default=[], choices=[ExcludedComponent.DDC, ExcludedComponent.Debug, ExcludedComponent.Templates], help='Exclude the specified component (can be specified multiple times to exclude multiple components)')
 		parser.add_argument('--opt', action='append', default=[], help='Set an advanced configuration option (can be specified multiple times to specify multiple options)')
 		parser.add_argument('--cuda', default=None, metavar='VERSION', help='Add CUDA support as well as OpenGL support when building Linux containers')
+		parser.add_argument('--visual-studio', default=VisualStudio.VS2017, choices=VisualStudio.BuildNumbers.keys(), help='Specify Visual Studio Build Tools version to use for Windows containers')
 		parser.add_argument('-username', default=None, help='Specify the username to use when cloning the git repository')
 		parser.add_argument('-password', default=None, help='Specify the password or access token to use when cloning the git repository')
 		parser.add_argument('-repo', default=None, help='Set the custom git repository to clone when "custom" is specified as the release value')
@@ -201,6 +219,8 @@ class BuildConfiguration(object):
 			'--build-arg', 'EXCLUDE_DEBUG={}'.format(1 if ExcludedComponent.Debug in self.excludedComponents else 0),
 			'--build-arg', 'EXCLUDE_TEMPLATES={}'.format(1 if ExcludedComponent.Templates in self.excludedComponents else 0)
 		]
+
+		self.uatBuildFlags = []
 		
 		# If we're building Windows containers, generate our Windows-specific configuration settings
 		if self.containerPlatform == 'windows':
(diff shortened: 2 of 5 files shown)


TBBle:
I think long-term it'd be better to have the vs version always in the tags, so there's never any doubt. Because the pre-reqs images are cross-version compatible, we can't really leave out the tag for a 'default' version of VS, because that would vary by release. On Linux, [the toolchain _is_ UE version-based](https://docs.unrealengine.com/4.26/en-US/SharingAndReleasing/Linux/NativeToolchain/), so we don't have that issue there.

I'd also suggest that this might be a good time to support parameterisation on VC version in a single channel, rather than switching channels, as I've described elsewhere: https://github.com/adamrehn/ue4-docker/issues/133#issuecomment-800115247. This would mean always using the current ("16" channel), but if we only want VC 14.16 (from VS2017) then pulling that from the "16" channel. This would provide the hook-site for users to specify (if necessary) a _specific_ VC build to include, and the rest of the inclusions are the same irrespective of channel. (Although based on your patch, we aren't using anything that is differently named in different channels right now.)

It'd also be useful, as you have here, to be able to say "2017" and "2019" (and eventually "2022") which would just pull the latest compiler for that release, `Microsoft.VisualStudio.Component.VC.v141.x86.x64` and `Microsoft.VisualStudio.Component.VC.Tools.x86.x64` respectively.

As an alternative thought, how painful would it be for VS install to be part of the ue4-source image instead, which would more-closely match the behaviour of the Linux toolchain, and make this question somewhat easier to give good defaults for, as we could default to the Epic-used version _for that build_, providing default ABI compatibility, as we get on Linux (by use of a fixed toolchain with specific sysroot).

slonopotamus:
> how painful would it be for VS install to be part of the ue4-source image instead

We would have to install VS twice then: in ue4-source and in ue4-minimal.

ue4docker/infrastructure/BuildConfiguration.py
```diff
@@ -226,15 +238,24 @@ def _generateWindowsConfig(self):
...
 		self.defaultDllDir = hostSysnative if os.path.exists(hostSysnative) else hostSystem32
 		self.dlldir = self.args.dlldir if self.args.dlldir is not None else self.defaultDllDir
-		
+
+		self.visualStudio = self.args.visual_studio
+
+		# Check whether specified Unreal Engine release is compatible with specified Visual Studio
+		vs_min_supported_unreal = VisualStudio.MinSupportedUnreal.get(self.visualStudio, None)
+		if vs_min_supported_unreal is not None and self.release < vs_min_supported_unreal:
+			raise RuntimeError('specified version of Unreal Engine cannot be built with Visual Studio 2019, oldest supported is {}'.format(vs_min_supported_unreal))
```
TBBle, line 247: "2019" here probably should be `self.visualStudio`, so that the error message doesn't have to change later.

slonopotamus: Right, I've overlooked this place when was refactoring things around.

ue4docker/infrastructure/BuildConfiguration.py
```diff
@@ -126,10 +139,9 @@ def __init__(self, parser, argv):
 			
 			# Validate the specified version string
 			try:
-				ue4Version = semver.parse(self.args.release)
-				if ue4Version['major'] != 4 or ue4Version['prerelease'] != None:
+				self.release = semver.VersionInfo.parse(self.args.release)
+				if self.release.major != 4 or self.release.prerelease is not None:
 					raise Exception()
-				self.release = semver.format_version(ue4Version['major'], ue4Version['minor'], ue4Version['patch'])
```
TBBle, line 132: This appears to have changed the format of `self.release`, but the same hasn't been done for the 'custom' side, which will do the wrong thing in the changed `_generateWindowsConfig`, I assume, as for custom builds that will be comparing a string `< vs_min_supported_unreal`.

Maybe if we need to actually track/mess with UE4 version for more than stringifying, I'd suggest we store that in a separate attribute, and either require the user to specify the base UE4 version when they do a custom-source build (not a bad idea, I think, it'd simplify other things), or somehow detect the base UE4 version, perhaps using `git fetch` to pull the Build.version file from the supplied repo.

TBBle:
> > how painful would it be for VS install to be part of the ue4-source image instead
> 
> We would have to install VS twice then: in ue4-source and in ue4-minimal.

How do we avoid that on Linux? We don't call setup.sh again in ue4-minimal, that I could see... Ah, I see. We explicitly [manually copy the toolchain from the builder image](https://github.com/adamrehn/ue4-docker/blob/615cc30d0c885a7646c78c4d576c7d84e8055a44/ue4docker/dockerfiles/ue4-minimal/linux/Dockerfile#L57-L61).

We probably can't do that for Windows, and I'm not sure implementing the moral equivalent (copying the downloaded install files from ue4-source into ue4-minimal and running the installer again) is worth the effort.

slonopotamus:
> I'd also suggest that this might be a good time to support parameterisation on VC version in a single channel, rather than switching channels, as I've described elsewhere: #133 (comment).

While this sounds like a good idea, I'd like to avoid expanding the scope of current PR to cover that. I'm not sure old UE versions support such setup (will they even be able to *find* the toolchain?), so it would need a lot of testing.

ue4docker/infrastructure/BuildConfiguration.py
slonopotamus: Yep, i broke 'custom' branch. Thanks for spotting. Fixed.

ue4docker/infrastructure/BuildConfiguration.py
slonopotamus: Fixed

slonopotamus:
Okay, this doesn't work as-is. VS2019 Build Tools doesn't have .NET SDK 4.5/4.6.2 anymore (I have no idea why installer doesn't exit with error and instead silently pretends as if everything happily installed).

image

Either we need to:
1. Install .NET SDK 4.5/4.6.2 by other means
2. Restrict VS2019 to UE 4.25+ that has https://github.com/EpicGames/UnrealEngine/commit/5256eedbdef30212ab69fdf4c09e898098959683 and can use whatever .NET SDK is present.

slonopotamus:
This also means that we can't easily go the "always use 16 channel" route.

TBBle:
I think in the [VS 2019 channel](https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2019#c-build-tools), you want `Microsoft.Net.ComponentGroup.4.6.2.DeveloperTools`, which according to [these docs](https://docs.microsoft.com/en-us/dotnet/framework/install/guide-for-developers#to-install-the-net-framework-developer-pack-or-targeting-pack) includes both the SDK and the targeting pack.

ue4docker/infrastructure/BuildConfiguration.py
```diff
@@ -226,15 +244,25 @@ def _generateWindowsConfig(self):
...
 		self.dlldir = self.args.dlldir if self.args.dlldir is not None else self.defaultDllDir
-		
+
+		self.visualStudio = self.args.visual_studio
+
+		if not self.custom:
+			# Check whether specified Unreal Engine release is compatible with specified Visual Studio
+			vsMinSupportedUnreal = VisualStudio.MinSupportedUnreal.get(self.visualStudio, None)
+			if vsMinSupportedUnreal is not None and semver.VersionInfo.parse(self.release) < vsMinSupportedUnreal:
+				raise RuntimeError('specified version of Unreal Engine cannot be built with Visual Studio {}}, oldest supported is {}'.format(self.visualStudio, vsMinSupportedUnreal))
```
TBBle, line 254: `{}}` probably should be `{}`?

ue4docker/dockerfiles/ue4-build-prerequisites/windows/install-prerequisites.bat
```diff
@@ -10,24 +10,27 @@ call refreshenv
...
 	--add Microsoft.VisualStudio.Workload.MSBuildTools ^
 	--add Microsoft.VisualStudio.Component.NuGet ^
 	--add Microsoft.VisualStudio.Component.Roslyn.Compiler ^
+	--add Microsoft.VisualStudio.Component.VC.Tools.x86.x64 ^
 	--add Microsoft.VisualStudio.Component.Windows10SDK.17763 ^
-	--add Microsoft.Net.Component.4.5.TargetingPack ^
-	--add Microsoft.Net.Component.4.6.2.SDK ^
-	--add Microsoft.Net.Component.4.6.2.TargetingPack
+	--add Microsoft.Net.ComponentGroup.4.6.2.DeveloperTools ^
+	--add Microsoft.Net.Component.4.5.TargetingPack
```
TBBle, line 32: I'm curious why we need the 4.5 Targeting Pack. (I realise it's not new, just changed lines, hence only curious). Particularly because I don't understand how a TargetingPack works without the relevant SDK installed as well.

ue4docker/infrastructure/BuildConfiguration.py
```diff
@@ -27,6 +27,23 @@
...
+	BuildNumbers = {
+		'2017' : '15',
+		'2019' : '16',
+	}
+
+	MinSupportedUnreal = {
+		# Unreal Engine 4.23.1 is the first that successfully builds with Visual Studio v16.3
+		# See https://github.com/EpicGames/UnrealEngine/commit/2510d4fd07a35ba5bff6ac2c7becaa6e8b7f11fa
+		#
+		# Unreal Engine 4.25 is the first that works with .NET SDK 4.7+
```
TBBle, line 43: With the current change, we still have the 4.6.2 SDK installed under VS2019, so I'd hope we can use VS 2019 against 4.23.1 now, if the user desires.

TBBle (approved):

ue4docker/dockerfiles/ue4-build-prerequisites/windows/install-prerequisites.bat
slonopotamus: I'm not a .NET expert, but from what I've read, Targeting Pack is something that contains .NET DLLs + some MSBuild config files so MSBuild understands how to build .NET applications against specified .NET.

And SDK is a different thing. Most noticeable artifacts it installs is C++ headers under `C:\Program Files (x86)\Windows Kits\NETFXSDK.

And now the weird bit: 
1. VS2017 has 4.6.2 SDK
2. VS2019 has 4.6.2 SDK
3. But VS2019 *Build Tools* doesn't have 4.6.2 SDK. When you ask it to install `Microsoft.Net.ComponentGroup.4.6.2.DeveloperTools`, it will actually install 4.6.2 Targeting Pack + *4.8* SDK. I'm not sure whether will will work at all, running a build currently.

TBBle: Interesting. The VS2019 Build Tools install is doing is what I would have _expected_: that the Targeting Pack for 4.N + _latest_ 4.x SDK would let you build targeting the .NET Framework 4.N, so you didn't have to keep multiple copies of the SDK around when each was a superset of the previous anyway.

That was my original vague understanding of .NET Framework development and the Targeting Packs, but I'd never actually had to put that to the test.

So perhaps Build Tools is right, but the UI installers let you be inefficient if you want by multi-installing SDKs or installing only old versions of the SDKs, and they've now trimmed that from the Build Tools installer as not necessary.

ue4docker/infrastructure/BuildConfiguration.py
slonopotamus: We can't (read other thread in this PR). We have 4.8 SDK under VS2019 Build Tools and older versions of UE4 are unable to use it.

I should possibly put a comment somewhere in the code about this situation.

ue4docker/dockerfiles/ue4-build-prerequisites/windows/install-prerequisites.bat
slonopotamus: > I'm curious why we need the 4.5 Targeting Pack

I don't remember actually. I discovered that it is required when was working on #144. Some of UE4 C# tools builds against 4.5, but I forgot which tool exactly.

slonopotamus:
Tested against: 4.20.3 + VS2017, 4.25.1 + VS2017, 4.25.1 + VS2019. 4.25.1 was used because it's the oldest version that is supposed to properly build with VS2019.

PR #186, opened 2021-08-16
Introduce codestyle via Black

slonopotamus:
Codestyle is a painful theme where lots of bikeshedding happens. I don't personally care what codestyle to use. I like when there is _a_ codestyle, whatever it is. And, importantly, when there's a tool that would automatically make my code compliant. This PR attempts to do that by using [Black](https://github.com/psf/black) formatter.


TBBle (approved):
100% agree for having a known and CI-validated code formatter. I use black elsewhere, so I also support this choice.

I am assuming that other that the .github/ folder, all the other changes are just the reformatting, so I didn't look at them. ^_^

TBBle:
Looking at this review made me realise we don't have any kind of "development environment" setup instructions. I guess `black` is the first thing that's a development-only concern, and so introducing that system is probably not yet important.

slonopotamus:
@adamrehn Are you OK with this?

adamrehn:
@slonopotamus I'm happy with this. Do you know if there's a way to trigger this automatically (e.g. via a pre-commit hook) or are we just going to rely on the linting CI step to warn when code hasn't been formatted?

jonpas:
You can do it via pre-commit, as long as user installs it.

PR #187, opened 2021-08-19
Use mcr.microsoft.com/windows image as a source for DLLs instead of host system

### LINKED ISSUE #165: Stop copying DLLs from host system and instead copy them from mcr.microsoft.com/windows image
Subj. Note that we will only be able to do this when we drop Windows Server 2016 LTSC support.

The benefit of this change is that we will be able to build ue4-build-prerequisites image even if host OS doesn't match target OS.

slonopotamus:
This commit:
1. Drops Windows 2016 LTSC support.
 Now the oldest supported Windows version is 2019 LTSC a.k.a. 1809
2. Stops copying DLLs from host system and instead uses mcr.microsoft.com/windows.
 This change allows building `ue4-build-prerequisites` even if host system version doesn't match.
 As a consequence of this change, `ue4-docker build` no longer has `-dlldir` command-line argument.
3. Extends list of copied DLLs to allow usage of Windows Advanced Rasterization Platform (WARP)
 in images derived from `ue4-build-prerequisites` image out of the box.

 Resolves #165

----

Not tested yet, I'm currently running builds against these changes.


ue4docker/dockerfiles/ue4-build-prerequisites/windows/Dockerfile
```diff
@@ -35,10 +39,22 @@ RUN start /wait %TEMP%\directx_redist.exe /Q /T:%TEMP% && `
...
+RUN curl --progress -L "https://github.com/microsoft/DirectXShaderCompiler/releases/download/v1.6.2104/dxc_2021_04-20.zip" --output %TEMP%\dxc.zip && `
+	powershell -Command "Expand-Archive -Path \"$env:TEMP\dxc.zip\" -DestinationPath $env:TEMP" && `
+	xcopy /y %TEMP%\bin\x64\dxcompiler.dll C:\GatheredDlls\ && `
+	xcopy /y %TEMP%\bin\x64\dxil.dll C:\GatheredDlls\
+
 # Gather the Vulkan runtime library
 RUN curl --progress-bar -L "https://sdk.lunarg.com/sdk/download/latest/windows/vulkan-runtime-components.zip?u=" --output %TEMP%\vulkan-runtime-components.zip
 RUN 7z e %TEMP%\vulkan-runtime-components.zip -oC:\GatheredDlls -y "*\x64\vulkan-1.dll"
 
+# Verify that the DLL files copied from the host can be loaded by the container OS
```
slonopotamus, line 52: I moved verification step to a later stage so that it covers DirectX and Vulkan libs.

slonopotamus:
Okay, I think that can be simplified even further by completely avoiding any work inside full Windows image and just using Docker `COPY` to transfer dlls directly from `C:\System32`. Will push updated changes tomorrow.

TBBle:
Perfect time to drop LTSC 2016 support, since [LTSC 2022 *just shipped*](https://docs.microsoft.com/en-us/windows-server/get-started/windows-server-release-info).

That said, this PR won't be enough to support it, because the source image is `mcr.microsoft.com/windows-server` rather than `mcr.microsoft.com/windows` as it is for LTSC 2019. But that can be a future issue, along with proper LTSC 2022 support/testing.

ue4docker/dockerfiles/ue4-build-prerequisites/windows/Dockerfile
```diff
@@ -1,54 +1,34 @@
...
 SHELL ["cmd", "/S", "/C"]
-COPY --from=dlls C:\GatheredDlls\ C:\Windows\System32\
 
 {% if not disable_labels %}
 # Add a sentinel label so we can easily identify all derived images, including intermediate images
 LABEL com.adamrehn.ue4-docker.sentinel="1"
 {% endif %}
 
+# Gather the system DLLs that we need from the full Windows base image
+COPY --from=dlls `
```
TBBle, line 14: You could literally use
```
COPY --from=mcr.microsoft.com/windows:${BASETAG}
```
here.

ue4docker/infrastructure/WindowsUtils.py
```diff
@@ -7,18 +7,21 @@
...
 
+	# The oldest Windows build we support
+	_minimumRequiredBuild = 17763
+
 	# The latest Windows build version we recognise as a non-Insider build
 	_latestReleaseBuild = 19042
 
 	# The list of Windows Server Core base image tags that we recognise, in ascending version number order
-	_validTags = ['ltsc2016', '1709', '1803', 'ltsc2019', '1903', '1909', '2004', '20H2']
+	_validTags = ['1809', '1903', '1909', '2004', '20H2']
```
TBBle, line 17: I think this change (`ltsc2019`->`1809` specifically) is the wrong direction. `ltsc2022` images don't have an equivalent SAC release tag, and it would be more consistent to refer to the LTSC releases by their official `ltscXXXX` tag, so we'll need to bring back management for that later anyway. I'm not sure what the `DisplayVersion` on Server LTSC 2022 is, it might be 21H2 (there was some references to that elsewhere) but the images produced by ue4-docker _should_ still say `ltsc2019` or `ltsc2022` respectively, since "1809" is weird if you don't realise that's "Server 2019", and _any_ SAC version for LTSC 2022 won't actually exist as a tag, or anywhere else in the collective consciousness of the users.

TBBle (commented):
I like the "drop LTSC 2016 support" part of this, and the "COPY --from=mcr.microsoft.com/..." part of this, but I believe the "call Server LTSC 2019 '1809'" change is a regression in the generated container tags, and also association with the target use-case.

ue4docker/dockerfiles/ue4-build-prerequisites/windows/Dockerfile
slonopotamus: Oh, cool. Didn't know about such feature.

ue4docker/infrastructure/WindowsUtils.py
slonopotamus: Uh... Okay. I just wanted to avoid ths back-and-forth mapping because there isn't `mcr.microsoft.com/windows:ltsc2019`.

TBBle: I remembered, the same-kernel release of Azure Stack HCI is 21H2, that's where I saw that numbering. It's still in Preview, so perhaps when it ships, perhaps there'll be 21H2-tagged container images.

However, I _suspect_ that the intended direction is for Azure Stack HCI to host LTSC 2022 VMs, and then _those_ host LTSC 2022-based containers orchestrated with Kubernetes. The Azure Stack HCI documentation is _really_ lean on such details though, so I'm guessing heavily there.

slonopotamus:
Okay, I got your point. Working on resurrecting `ltsc2019` tag...

ue4docker/infrastructure/WindowsUtils.py
TBBle: Yeah, lack of `mcr.microsoft.com/windows:ltsc2019` is annoying, I hadn't noticed that.

I think we'll have to manually handle that, because it's a unique case (the other SAC/Client versions we support do have equivalents there), and going forward, no new tags will be added to that repo anyway; instead we'll need to support `mcr.microsoft.com/windows/server` which _only_ has `ltsc2022` and kernel-version tags, and I expect will never have SAC-style tags.

ue4docker/dockerfiles/ue4-build-prerequisites/windows/Dockerfile
TBBle: Yeah, I only noticed it recently, while helping a colleague implement a *suspiciously similar* DLL copying routine for a new CI image, and noticed such an example in the Dockerfile reference documentation.

slonopotamus: Hmm... That syntax actually doesn't work for me. See https://github.com/moby/moby/issues/34482

slonopotamus: So it looks like I cannot `COPY --from ${var}`.

slonopotamus:
I've restored `ltsc2019` tag behavior. And yep, we now need to ugly functions to map back and forth: https://github.com/adamrehn/ue4-docker/blob/5ca088ed52a3dbe8808dae402b774b7c51753d8f/ue4docker/infrastructure/WindowsUtils.py#L118-L139

The second function will be additionally used to adjust full image name for ltsc2022 (that's why it returns full image name instead of just tag).

ue4docker/build.py
```diff
@@ -245,7 +231,7 @@ def build():
 			prereqsArgs = ['--build-arg', 'BASEIMAGE=' + config.baseImage]
 			if config.containerPlatform == 'windows':
 				prereqsArgs = prereqsArgs + [
-					'--build-arg', 'HOST_BUILD=' + str(WindowsUtils.getWindowsBuild()),
+					'--build-arg', 'FULLBASEIMAGE=' + config.fullBaseImage,
```
slonopotamus, line 234: I'm not 100% happy with naming of this thing, but failed to come up with a better version. We can't use "desktop image" term because `https://mcr.microsoft.com/windows/server` isn't desktop in any way.

ue4docker/dockerfiles/ue4-build-prerequisites/windows/Dockerfile
TBBle: Ah, I see. Yeah, [this blog post](https://medium.com/@tonistiigi/advanced-multi-stage-build-patterns-6f741b852fae) explains that they rely on `FROM` to evaluate args in image references in order to determine the required input images, and that issue discussion suggests the maintainers felt that treating the `--from=` the same way as `FROM` would be confusing, since it would be different to how other things in that line are evaluated.

It's a shame, but it's not unreadable. The definition of `dlls` isn't right next to where it's used, but that's not the worst possible outcome. I kind-of wish we had `IMAGE`, which would be `FROM` but without declaring a new stage, aka `ARG` for images.

But we don't have BuildKit on Windows yet, let alone scope to experiment with the language. ^_^

TBBle (approved):

ue4docker/build.py
TBBle: `DLLSRCIMAGE` perhaps? Since we never run it, there's nothing preventing someone using an image that contains _just_ the desired DLLs extracted elsewhere, for example.

ue4docker/infrastructure/WindowsUtils.py
slonopotamus: > our target audience won't know what 21H2 is, or worse, will confuse it with Windows 10 21H2

I'm not sure I understand what you mean here. ltsc2022 identifies itself as 21H2, and MS clearly says that: https://docs.microsoft.com/en-us/windows-server/get-started/windows-server-release-info#windows-server-current-versions-by-servicing-option

TBBle: Ah, I hadn't noticed they were using 21H2 in the official docs. They're not using it in the Docker tags though, and I _still_ think it's confusing given that [Windows 10 21H2 also exists](https://blogs.windows.com/windowsexperience/2021/07/15/introducing-the-next-feature-update-to-windows-10-21h2/), and is running a different kernel version. Apparently [Windows 11 previews are _also_ 21H2](https://techcommunity.microsoft.com/t5/windows-hardware-certification/now-accepting-windows-11-version-21h2-submissions/ba-p/2483049), so it's pretty much out-the-window as a shorthand for any particular Kernel Build, which is what we actually care about in container-land.

And for ue4-docker, if they are using 21H2 as `DisplayVersion`, we'll need to check elsewhere in that key to distinguish Windows Server 2022 from Windows 10 21H2, as we can only support the former with process isolation, barring some kind of unexpected turnaround on current OS container image plans.

The SAC version numbers are a neat shorthand, but I think using them for releases without a kernel-compatible SAC release is probably a mistake on MS's part, and going to lead to confusion over time. I assume this is why they're not using `21h2` as a container image tag, and `ltsc2022` is now primary (based on it having the KB variants attached, while `ltsc2019` did not).

I kind-of hope they "disappear" `21H2` in reference to Windows Sever 2022 pretty quickly. I still think that long-term, we will need to drop the SAC version names in favour of matching the actual build numbers, but I recall that didn't shake out well last time you tried it.

slonopotamus: > Windows 10 21H2 also exists, and is running a different kernel version. Apparently Windows 11 will also be 21H2

Oh my god, what a mess.

> we will need to drop the SAC version names and ltsc floating tags in favour of matching the actual build numbers, but I recall that didn't shake out well last time you tried it.

Yep, I didn't like what it turned into. Possibly needs another attempt given that we now have more info about future Windows versioning.

ue4docker/build.py
slonopotamus: `DLLSRCIMAGE` is good, thanks for suggestion.

ue4docker/infrastructure/WindowsUtils.py
slonopotamus: See https://github.com/adamrehn/ue4-docker/pull/191

PR #191, opened 2021-08-21
Stop relying on ReleaseId/DisplayVersion registry keys

slonopotamus:
Instead, we now use Windows build numbers to identify host system.

This is a preparation step to add support for Windows Server 2022 LTSC,
 whose DisplayVersion (21H2) collides with Windows 10 21H2 and Windows 11 21H2.

This commit also removes EOLed Windows 1903 and 1909 in order to reduce lookup tables that we have to hardcode.

### DIFF
--- a/ue4docker/infrastructure/WindowsUtils.py
+++ b/ue4docker/infrastructure/WindowsUtils.py
@@ -1,6 +1,7 @@
 from .DockerUtils import DockerUtils
 from pkg_resources import parse_version
 import platform, sys
+from typing import Optional
 
 if platform.system() == 'Windows':
 	import winreg
@@ -10,18 +11,22 @@ class WindowsUtils(object):
 	# The oldest Windows build we support
 	_minimumRequiredBuild = 17763
 
-	# The latest Windows build version we recognise as a non-Insider build
-	_latestReleaseBuild = 19042
+	# This lookup table is based on the list of valid tags from <https://hub.docker.com/r/microsoft/windowsservercore/>
+	# and list of build-to-release mapping from https://docs.microsoft.com/en-us/windows/release-health/release-information
+	_knownTagsByBuildNumber = {
+		17763: 'ltsc2019',
+		18362: '1903',
+		18363: '1909',
+		19041: '2004',
+		19042: '20H2',
+		19043: '21H1',
+	}
 
-	# The list of Windows Server Core base image tags that we recognise, in ascending version number order
-	_validTags = ['ltsc2019', '1903', '1909', '2004', '20H2']
+	_knownTags = list(_knownTagsByBuildNumber.values())
 
 	# The list of Windows Server and Windows 10 host OS releases that are blacklisted due to critical bugs
 	# (See: <https://unrealcontainers.com/docs/concepts/windows-containers>)
-	_blacklistedReleases = ['1903', '1909']
-
-	# The list of Windows Server Core container image releases that are unsupported due to having reached EOL
-	_eolReleases = ['1903', '1909']
+	_blacklistedHosts = [18362, 18363]
 
 	@staticmethod
 	def _getVersionRegKey(subkey : str) -> str:
@@ -54,25 +59,21 @@ class WindowsUtils(object):
 		'''
 		Generates a verbose human-readable version string for the Windows host system
 		'''
-		return '{} Version {} (Build {}.{})'.format(
+		return '{} (Build {}.{})'.format(
 			WindowsUtils._getVersionRegKey('ProductName'),
-			WindowsUtils.getWindowsRelease(),
 			WindowsUtils.getWindowsBuild(),
 			WindowsUtils._getVersionRegKey('UBR')
 		)
 
 	@staticmethod
-	def getWindowsRelease() -> str:
+	def getHostBaseTag() -> Optional[str]:
 		'''
-		Determines the Windows 10 / Windows Server release (1607, 1709, 1803, etc.) of the Windows host system
+		Retrieves the tag for the Windows Server Core base image matching the host Windows system
 		'''
-		try:
-			# Starting with Windows 20H2 (also known as 2009), Microsoft stopped updating ReleaseId
-			# and instead updates DisplayVersion
-			return WindowsUtils._getVersionRegKey('DisplayVersion')
-		except FileNotFoundError:
-			# Fallback to ReleaseId for pre-20H2 releases that didn't have DisplayVersion
-			return WindowsUtils._getVersionRegKey('ReleaseId')
+
+		hostBuild = WindowsUtils.getWindowsBuild()
+
+		return WindowsUtils._knownTagsByBuildNumber.get(hostBuild)
 
 	@staticmethod
 	def getWindowsBuild() -> int:
@@ -82,23 +83,14 @@ class WindowsUtils(object):
 		return sys.getwindowsversion().build
 
 	@staticmethod
-	def isBlacklistedWindowsVersion(release=None):
+	def isBlacklistedWindowsHost() -> bool:
 		'''
-		Determines if the specified Windows release is one with bugs that make it unsuitable for use
+		Determines if host Windows version is one with bugs that make it unsuitable for use
 		(defaults to checking the host OS release if one is not specified)
 		'''
 		dockerVersion = parse_version(DockerUtils.version()['Version'])
-		release = WindowsUtils.getWindowsRelease() if release is None else release
-		return release in WindowsUtils._blacklistedReleases and dockerVersion < parse_version('19.03.6')
-
-	@staticmethod
-	def isEndOfLifeWindowsVersion(release=None):
-		'''
-		Determines if the specified Windows release is one that has reached End Of Life (EOL)
-		(defaults to checking the host OS release if one is not specified)
-		'''
-		release = WindowsUtils.getWindowsRelease() if release is None else release
-		return release in WindowsUtils._eolReleases
+		build = WindowsUtils.getWindowsBuild()
+		return build in WindowsUtils._blacklistedHosts and dockerVersion < parse_version('19.03.6')
 
 	@staticmethod
 	def isWindowsServer() -> bool:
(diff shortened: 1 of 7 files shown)


ue4docker/infrastructure/WindowsUtils.py
```diff
@@ -116,19 +98,18 @@ def isInsiderPreview() -> bool:
...
 
-		# This lookup table is based on the list of valid tags from <https://hub.docker.com/r/microsoft/windowsservercore/>
-		return {
-			'1809': 'ltsc2019',
-		}.get(release, release)
+		# Note that we fallback to hostBuild to avoid pretending that everything is fine
+		# if we're running on an unsupported Windows build (like, EOLed or Insider Preview).
+		# This fallback will later be rejected when ue4-docker checks whether it is valid.
+		# Users of unsupported Windows builds can still use ue4-docker by using -basetag option with a known value.
+		return WindowsUtils._tagsByBuildNumber.get(hostBuild, str(hostBuild))
```
slonopotamus, line 112: @TBBle This particular change is a bit questionable, maybe you have some comments on it?

TBBle: Perhaps we could just return `None` and catch it elsewhere, requiring that the user provide an explicit base-tag (and possibly force process isolation if they know it's safe) They'd still get warnings, which seems appropriate for "You have to know what you're doing at this point".

Alternatively, we could return the full version (10.0.XXXX.YY) which would be a valid tag if there's a matching image published (should be the case for all non-Insider builds), and hence is somewhat future-proof, but might be surprising to users who would have expected failure in that case, and certainly would have expected a floating tag (SAC-style or `ltsc20xx`) for the resulting image.

ue4docker/infrastructure/WindowsUtils.py
```diff
@@ -10,18 +10,20 @@ class WindowsUtils(object):
...
 	# The list of Windows Server Core base image tags that we recognise, in ascending version number order
-	_validTags = ['ltsc2019', '1903', '1909', '2004', '20H2']
+	_validTags = list(_tagsByBuildNumber.values())
 
 	# The list of Windows Server and Windows 10 host OS releases that are blacklisted due to critical bugs
 	# (See: <https://unrealcontainers.com/docs/concepts/windows-containers>)
-	_blacklistedReleases = ['1903', '1909']
-
-	# The list of Windows Server Core container image releases that are unsupported due to having reached EOL
-	_eolReleases = ['1903', '1909']
```
TBBle, line 24: Windows 10 2004 hits EOL in December, and 20H2 EOLs in May next year. So if we just disappear them entirely, then we lose support for any Windows Client platforms.

I'd keep the EOL support around until we either decide to drop Windows 10 process isolation support once there's no supported base images available, or something changes on the MS side, or some other intervening idea occurs.

Better to say "Yes, we recognise this, it's EOL, are you sure you want to continue" when we can recognise a version.

Also, these two versions are only blacklisted for Docker versions before 19.03.6, so this change actually _promotes_ them from EOL warning with user-override to "supported", i.e. silently accepted.

slonopotamus: The trick is that they're blacklisted as a _host_ and EOL as a _target_. We silently allow building supported versions on EOL hosts.

slonopotamus:
I think I broke the logic that checks if host is older than target _again_ (as it attempted to happen in #166).

ue4docker/infrastructure/WindowsUtils.py
slonopotamus: > Windows 10 2004 hits EOL in December, and 20H2 EOLs in May next year. So if we just disappear them entirely, then we lose support for any Windows Client platforms.

OK, agreed.

TBBle: Ah, good point. I'm fine with not complaining about using an EOL host to build a supported target.

I'm ambivalent about complaining when we try to use an EOL _target_; I'd lean towards keeping the warning around based on the calculated base tag (i.e. _after_ applying the user's specified base tag), but that isn't really comprehensive or reliable (e.g., basetag `1909-KB4574727-amd64` would not be recognised), so I don't mind if we drop it now.

My personal use-cases are 100% process-isolation-based (and Windows Server LTSC-based 95% of the time), so I haven't had any value from the EOL warning myself.

I'd be interested to know if @adamrehn has any concerns about dropping this (or use-cases where it prevents a bad outcome), but if not, then I'm happy to resolve this thread.

TBBle:
> I think I broke the logic that checks if host is older than target _again_ (as it attempted to happen in #166).

Does making `_tagsByBuildNumber` an OrderedDict resolve it? At a brief glance, `isNewerBaseTag` depends on `_validTags` being ordered.

This whole test relies _heavily_ on `basetag<=>Build` equivalence, but unless we extract the Windows OS version from the manifest of the base container image, we have no other way of being sure that a given base tag is for an older, newer, or matching kernel version compared to the host.

ue4docker/infrastructure/WindowsUtils.py
slonopotamus: > basetag 1909-KB4574727-amd64 would not be recognised

ue4-docker doesn't support building targets that are not hardcoded in valid tags list _at all_. #166 wanted to change that, but kinda failed.

----

I thought about 20H2 again. I _very_ doubt that anyone would want to build ue4-docker images for it after it gets EOLed. If they're on a newer host, they would need to use Hyper-V isolation. And if you use Hyper-V isolation, why not just use a supported target version? Alternatively, they still run EOLed 20H2 on their host. Well, if you use EOLed Windows, also use EOLed version of ue4-docker that has 20H2 support, no?

slonopotamus:
`isNewerBaseTag` doesn't properly work for unknown things.

> or skipped entirely... If the user provides a base tag, we have to drop a lot of assumptions about the format of that tag).

Aaand... That's exactly what I did in #166 :D

TBBle:
And as I hid in my last wall-of-text comment on #166, I think sorting/ordering is fighting a losing battle. We can catch _some_ of the problems, _some_ of the time. We cannot correctly handle unknown things by SAC-style tags (i.e. `ReleaseVersion` registry value) because Windows Client 21H2 is both newer (Windows 11: 22000) and older (Windows 10: 19044) than Windows Server 21H2 (20348). That said, none of those three exist as images with a 21H2 tag, and this is a new situation compared to last time we looked at this, but I think that just strengthens my idea that we can't handle unknown things, because we definitely would have guessed wrong last time.

We can know the host build number, but for base tags, we can only know by either inspecting the manifest (which assumes we _have_ the manifest at the time of running ue4-docker, and we can't assume that), or by knowing that certain images (path + tag) map to certain kernel builds, and that goes out the window as soon as I locally modify ue4-docker to use a local pull-through on-site Artifactory proxy in front of mcr.microsoft.com (or to target Insider builds, or _both_), or need to target a specific immutable version tag with `-basetag`, e.g. because I need a base image before the Feb 2020 security update to run on a not-updated LTSC 2019 server.

slonopotamus:
So, let's agree that we want to extend current "we only accept things that we know" to "we accept whatever user gave as input" and just disable our checks for unknown tags?

adamrehn:
Regarding version ordering, there are really only two scenarios where we actually care about the version relationship between the host and the target:

- When copying DLL files from the host system (no longer an issue after #187.)

- When determining whether the host is capable of running target containers via either process isolation or Hyper-V isolation.

All of the ordering checks were implemented with DLL copying in mind and they've resulted in a number of limitations, such as awkwardness when using Insider Builds of Windows. With the DLL copying issue resolved, I'm all in favour of eliminating the ordering checks and performing a simple test to verify the ability to run containers as described by @TBBle above.

TBBle:
> We could try guessing, and perhaps even inspect the image (I _think_ we can get that metadata) but that's a bonus at that point.

I checked in-passing, and `docker image inspect` will include the Windows build version the image is targeting. So if we go this route, then once we've `docker run mcr...` and it's failed, we can inspect that `mcr...` and hence have the relevant build numbers available.

We could in-fact auto-switch to Hyper-V isolation and/or reject the image if we first `pull` the base image, and then inspect it for the relevant metadata (I think it's marked as 'variant' in the Docker image metadata) before proceeding further.

Obviously if we're not going to actually _do_ the build now, we shouldn't be pulling or inspecting stuff, as the host system may not be the final place it's being run anyway.

Anyway, it's a future thing to do, but at least it seems feasible.

slonopotamus:
Okay, I think I did it this time.

We now:
1. Do not touch `ReleaseId`/`DisplayVersion` at all and use lookup table based on host build number.
2. We allow usage of *any* servercore base tag. The only caveat is that we skip checking whether that tag is newer/older and just issue a warning.

slonopotamus:
@TBBle did you have a chance to look at updated changes?

ue4docker/infrastructure/WindowsUtils.py
```diff
@@ -142,22 +112,18 @@ def getDllSrcImage(basetag: str) -> str:
...
 		'''
-		Determines if the specified tag is a valid Windows Server Core base image tag
+		Returns the list of known tags for the Windows Server Core base image, in ascending chronological release order
 		'''
-		return tag in WindowsUtils._validTags
+		return WindowsUtils._knownTags
 
 	@staticmethod
-	def isNewerBaseTag(older: str, newer: str) -> bool:
+	def isNewerBaseTag(older: str, newer: str) -> Optional[bool]:
```
TBBle, line 122: Non-blocking nit-pick: I'm reasonably unfond of `Optional[bool]`, but don't have a better option, except letting the `ValueError` bubble up, or throwing a custom exception, and catching that in the caller.

TBBle (approved):
Sorry, I accidentally marked this read, forgot I had an outstanding review.

ue4docker/infrastructure/WindowsUtils.py
slonopotamus: Dunno, these both look equally ugly to me :D

I'd actually make a enum { Newer, Older, Unknown } but don't think it will improve anything at all given that we're in Python that ~doesn't even have switch/case~. Okay, it got it in 3.10, but that's too much new yet to use.

PR #199, opened 2021-09-07
Add support for Windows Server 2022 LTSC

### LINKED ISSUE #190: Add support for Windows Server 2022
Subj.

slonopotamus:
resolves #190

----

Testing in progress...


TBBle (approved):

TBBle:
"Custom base tag" is a simpler variant of the "Insider" problem, so `-dllimage` seems reasonable. We could use `-baseimage` for "Insider", which also would support using a prepared base image from a local repo that builds on server core, as an alternative to building on top of the final images for site-local requirements.

Shouldn't block this change though.

slonopotamus:
Tested this against 4.20.3, looks fine.

PR #221, opened 2022-02-16
Add support for specifying linux basetag

### LINKED ISSUE #183: Add Ubuntu 20.04 as base image
Subj

russkel:
I had to use Ubuntu focal for library reasons and found there were no PRs for this support. I have built a `4.27.2-cudagl11.4` image with this which works well.

You can now specify `-basetag=ubuntu20.04`.

Fixes #183

### DIFF
--- a/ue4docker/infrastructure/BuildConfiguration.py
+++ b/ue4docker/infrastructure/BuildConfiguration.py
@@ -14,15 +14,18 @@ DEFAULT_GIT_REPO = "https://github.com/EpicGames/UnrealEngine.git"
 
 # The base images for Linux containers
 LINUX_BASE_IMAGES = {
-    "opengl": "nvidia/opengl:1.0-glvnd-devel-ubuntu18.04",
+    "opengl": "nvidia/opengl:1.0-glvnd-devel-{ubuntu}",
     "cudagl": {
-        "9.2": "nvidia/cudagl:9.2-devel-ubuntu18.04",
-        "10.0": "nvidia/cudagl:10.0-devel-ubuntu18.04",
-        "10.1": "nvidia/cudagl:10.1-devel-ubuntu18.04",
-        "10.2": "nvidia/cudagl:10.2-devel-ubuntu18.04",
-        "11.4": "nvidia/cudagl:11.4.2-devel-ubuntu20.04",
+        "9.2": "nvidia/cudagl:9.2-devel-{ubuntu}",
+        "10.0": "nvidia/cudagl:10.0-devel-{ubuntu}",
+        "10.1": "nvidia/cudagl:10.1-devel-{ubuntu}",
+        "10.2": "nvidia/cudagl:10.2-devel-{ubuntu}",
+        "11.4": "nvidia/cudagl:11.4.2-devel-{ubuntu}",
     },
 }
+
+# The default ubuntu base to use
+DEFAULT_LINUX_VERSION = "ubuntu18.04"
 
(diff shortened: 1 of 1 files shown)


slonopotamus:
This is not quite how I imagined this feature... But, why not. Could you just please fix code formatting by [running Black](https://github.com/adamrehn/ue4-docker/blob/master/.github/CONTRIBUTING.md#autoformatting-your-changes)?

slonopotamus:
One more thought: on Windows, we use different tags for ue4-build-prerequisites images based on OS version inside it. Do we want to do the same on Linux?

@TBBle WDYT?

TBBle:
My leaning would be to not build the base image tag into the output tags if it doesn't indicate a hard compatability-break. But I guess changing Ubuntu version is as-breaking (or as non-breaking as you prefer) as changing Visual Studio version, since most of the time it won't matter, but when it does matter, having both base images distinctly in your registry might be valuable.

As a counter-thought, do we actually need to keep Ubuntu base version switchable? LTS is nice, but 4 years and with less than three months until it's _two_ LTS's behind, perhaps we can just switch to 20.04 base always?

Disclaimer, I'm not doing much Linux UE4 at the moment (or Linux containers in general), so maybe this is a terrible idea. Ubuntu 18.04 support expires in April 2023, apparently, so it's still valid to keep supporting it from that perspective.

russkel:
> As a counter-thought, do we actually need to keep Ubuntu base version switchable? LTS is nice, but 4 years and with less than three months until it's _two_ LTS's behind, "long term" is starting to stretch out to "obsolete". Perhaps we can just switch to 20.04 base always?

I thought there must have been a reason people were using 18.04 over the 20.04 LTS, hence it still being the base on this and Epic official images four years down the track. I think 18.04 is one of the worst combinations of software versions known to man and just simply doesn't work if you're doing even average complexity development without reinstalling 50% of the software packages.

slonopotamus:
> I thought there must have been a reason people were using 18.04 over the 20.04 LTS

The reason is trivial: 20.04 didn't exist at all when ue4-docker was born.

> Perhaps we can just switch to 20.04 base always?

The question is whether `Setup.sh` from older UE releases can handle that. If it can, I +1 on "just upgrade to 20.04".

russkel:
Thanks for merging. Sorry I didn't get around to linting.

adamrehn:
> @slonopotamus: The question is whether Setup.sh from older UE releases can handle that. If it can, I +1 on "just upgrade to 20.04". If they don't... We'll have to drop support for them. Maybe not now, but when 18.04 EOLs.

This only matters for Unreal Engine 4.20.3 and older, since 4.21.0 was the point when Mono started being bundled alongside clang and `Setup.sh` no longer needed to install any packages [other than build-essential](https://github.com/EpicGames/UnrealEngine/blob/4.21.0-release/Engine/Build/BatchFiles/Linux/Setup.sh#L26-L30). It's hard to know without actually testing it whether attempting to build Unreal Engine 4.20 under Ubuntu 22.04 would break things, since that version of `Setup.sh` would simply [treat unknown versions of Ubuntu as though they were Ubuntu 17.10](https://github.com/EpicGames/UnrealEngine/blob/4.20.3-release/Engine/Build/BatchFiles/Linux/Setup.sh#L99) and attempt to install packages using their names from the 17.10 repositories. If the package names and contents are unchanged in Ubuntu 22.04 then in theory it should still work just fine. If not, then dropping support for Unreal Engine 4.20 is an easy solution since all subsequent versions should work under any version of Ubuntu.

PR #222, opened 2022-02-17
Improve ue4-build-prerequisites on Linux

slonopotamus:
This is a follow-up to adamrehn/ue4-docker#221

This commit:
1. Removes hardcoded list of CUDA images. Now user can specify whatever version they want, and we do not need to maintain this list
2. Adds base OS name to ue4-build-prerequisites tag for clarity
3. Prints base image tag in `ue4 build` header output

----

@russkel I've somewhat extended your changes from #221, possibly you want to comment on this.

### DIFF
--- a/ue4docker/build.py
+++ b/ue4docker/build.py
@@ -262,18 +262,9 @@ def build():
                 sys.exit(1)
 
         elif config.containerPlatform == "linux":
-
-            # Determine if we are building CUDA-enabled container images
-            capabilities = (
-                "CUDA {} + OpenGL".format(config.cuda)
-                if config.cuda is not None
-                else "OpenGL"
-            )
             logger.info("LINUX CONTAINER SETTINGS", False)
             logger.info(
-                "Building GPU-enabled images compatible with NVIDIA Docker ({} support).\n".format(
-                    capabilities
-                ),
+                "Base OS image: {}\n".format(config.baseImage),
                 False,
             )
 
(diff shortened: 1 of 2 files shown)


slonopotamus:
The next logical step would be to completely remove `--cuda` flag and just use `-basetag` to specify full image name.

russkel:
Thanks for merging!

russkel:
> 1. Removes hardcoded list of CUDA images. Now user can specify whatever version they want, and we do not need to maintain this list

I think this makes sense. A simple query can check the existence of the image.
Also using an up-to-date default image that maybe tracks the latest version on Dockerhub?

slonopotamus:
> Also using an up-to-date default image that maybe tracks the latest version on Dockerhub?

I'd keep that out of scope of current PR.

adamrehn:
This seems reasonable to me. There have been a lot of *"add CUDA version X.YZ to the base images list"* requests over the years, so allowing users to specify an arbitrary version will definitely make things easier for end users and for the maintainers team.

With regards to using `-basetag` to specify a full image name, I think we could add this in a fairly backwards-compatible manner, by detecting the presence of a colon character to indicate a fully-qualified tag name and treating the absence of a colon as an indication that we should use the current behaviour where we fill out a predefined tag with an Ubuntu version number. This is similar to [the way that the `ue4-docker export` command behaves](https://github.com/adamrehn/ue4-docker/blob/v0.0.96/ue4docker/export.py#L63-L66) in terms of how users can specify image tags. (We could alternatively check for both a forward slash and a colon, but that would complicate scenarios where the user wants to specify an official image from Docker Hub, since they'd need to specify `library/ubuntu:22.04` instead of just `ubuntu:22.04` for example.)

In terms of removing the `--cuda` flag, my preference would be to retain it for backwards compatibility reasons, but to emit an error if it is specified in conjunction with a fully-qualified image tag.

PR #223, opened 2022-02-23
Add missing dependency for WebBrowserWidget plugin under Linux

butterman0:


russkel:
```
↳ ldd Engine/Binaries/ThirdParty/CEF3/Linux/libcef.so | grep damage
	libXdamage.so.1 => /lib/x86_64-linux-gnu/libXdamage.so.1 (0x00007fe76a3ae000)
```

PR #229, opened 2022-03-11
RFC: Use BuildKit for all images

slonopotamus:
BuildKit offers much shorter build times

----

This is a somewhat controversial change. On one hand, BuildKit severely improves build times. On the other hand, there *are* reasons why BuildKit is still not enabled by default in Docker.

Some numbers:

Linux
HDD
Ryzen 3700X
ue4-minimal only (ue4-source is already pre-built)
4.20.3
Flags: `--exclude debug --exclude ddc --no-engine --no-full`

`DOCKER_BUILDKIT=1` 1h 6m
`DOCKER_BUILDKIT=0` 1h 27m

### DIFF
--- a/ue4docker/infrastructure/ImageBuilder.py
+++ b/ue4docker/infrastructure/ImageBuilder.py
@@ -71,7 +71,7 @@ class ImageBuilder(object):
             # Determine whether we are building using `docker buildx` with build secrets
             imageTags = self._formatTags(name, tags)
             command = DockerUtils.build(imageTags, self.context(name), args)
-            env = None
+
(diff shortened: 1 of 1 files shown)


TBBle:
Since we turn it on magically in some circumstance, it makes sense to me to enable it always.

We lose the ability to diagnose build issues using the intermediate layers as images, but I guess that's mostly a tool we've used on WIndows due to filesize issues, where this won't work until we have BuildKit for Windows Containers anyway.

slonopotamus:
Okay, let's merge then. If we later understand that this makes us/users less happy, we can always revert this.

PR #234, opened 2022-04-04
Implement --target to replace the old --no-* options.

### LINKED ISSUE #227: `ue4-build-prerequisites`-only mode
Feature request, with offer of willingness to do work.

I'm working on a CI system built partly on ue4-docker. In our case, the source code is included in a p4 repo which is synced as part of the CI process and should not be built into the Docker image. Basically, we want `build-essentials` and nothing else, everything downstream of that is our problem. In our local branch of ue4-docker we've added a `--no-source` commandline option to suppress building the source image, and I'm happy to give you a pull request for it if you'd like.

But it's pretty dang janky to actually use.

The problem is that the `build` parameters are designed for `ue4-source` to *always* be generated, and therefore it expects, as the first token and without a named `--option`, either a UE4 version or the keyword "custom". Whatever UE4 version you provide is kind of intrinsically incorrect because, y'know, custom from the repo, whereas if you enter "custom" it then demands repo information (which is complicated to enter and also completely ignored because we never generate `ue4-source`). I've bypassed this in my local usage by simply passing in `4.0.0`, which surprisingly works, but, again, pretty dang janky.

It's not clear how to fix this. I think arguably this never should have been a bare argument in the first place, but I've certainly made that mistake in the past myself, and what's done is done. Perhaps this should be a separate build mode entirely (`generate-build-essentials`?) but then you're looking at moving a ton of code to be shared between `build.py` and `generate-build-essentials.py`. Perhaps this *should* be turned into a named argument, with a temporary compatibility mode where it also accepts it as a bare argument. Perhaps the code should just allow it to be skipped if `--no-source` is provided (which is weird in its own right because it means the arguments can't be parsed in order, unless `--no-source` has to be at the beginning, which is *also* weird) or perhaps there should be a special token for "I do not care about this, go away".

I'm happy to spend a little time implementing whatever seems reasonable as long as it's not a massive refactoring. Let me know what you'd prefer, and if you just want `--no-source` in its current format, janky ignored version number and all, I'll turn it into a pull request for ya :)

zorbathut:
Resolves #227.

This is undertested, I'm afraid; I don't have access to all platforms, nor do I have things set up to even build the `source` image. It is, however, perfectly happy to spit out layouts for all the images, and as I haven't touched any of the actual image generation code, it *should* work. But please test it! I'm assuming you have a functional framework for this.

I took some liberty in moving things into image-specific conditionals, mostly with the goal of making `--target=build-prerequisites` work. Most notably, main tag naming, credential server startup, and a considerable amount of parameter validation are now hidden under `buildTargetSource` checks. Unfortunately this makes the diff look a lot uglier than it should.

Let me know if you'd like changes!

### DIFF
--- a/ue4docker/build.py
+++ b/ue4docker/build.py
@@ -105,11 +105,12 @@ def build():
             config.combine,
         )
 
-        # Resolve our main set of tags for the generated images
-        mainTags = [
-            "{}{}-{}".format(config.release, config.suffix, config.prereqsTag),
-            config.release + config.suffix,
-        ]
+        # Resolve our main set of tags for the generated images; this is used only for Source and downstream
+        if config.buildTargetSource:
+            mainTags = [
+                "{}{}-{}".format(config.release, config.suffix, config.prereqsTag),
+                config.release + config.suffix,
+            ]
 
         # Print the command-line invocation that triggered this build, masking any supplied passwords
         args = [
@@ -128,10 +129,11 @@ def build():
         )
         if config.custom == True:
             logger.info("Custom name:   " + config.release, False)
-        else:
+        elif config.release is not None:
             logger.info("Release:       " + config.release, False)
-        logger.info("Repository:    " + config.repository, False)
-        logger.info("Branch/tag:    " + config.branch + "\n", False)
+        if config.repository is not None:
+            logger.info("Repository:    " + config.repository, False)
+            logger.info("Branch/tag:    " + config.branch + "\n", False)
 
         # Determine if we are using a custom version for ue4cli or conan-ue4cli
         if config.ue4cliVersion is not None or config.conanUe4cliVersion is not None:
@@ -241,7 +243,7 @@ def build():
 
             # Ensure the Docker daemon is configured correctly
             requiredLimit = WindowsUtils.requiredSizeLimit()
-            if DockerUtils.maxsize() < requiredLimit:
+            if DockerUtils.maxsize() < requiredLimit and config.buildTargetSource:
                 logger.error("SETUP REQUIRED:")
                 logger.error(
                     "The max image size for Windows containers must be set to at least {}GB.".format(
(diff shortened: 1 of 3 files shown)


slonopotamus:
1. We possibly want to print list of targets that will be built in initial `ue4-build` banner.
2. It looks like you're adding the same targets multiple times to `self.args.target` array when evaluating dependencies and converting from old-style to new-style. Maybe there should be a `set` somewhere?

ue4docker/build.py
```diff
@@ -363,91 +365,99 @@ def build():
...
+            else:
+                logger.info("Skipping ue4-source image build.")
 
             # Build the UE4 Engine source build image, unless requested otherwise by the user
-            ue4BuildArgs = prereqConsumerArgs + [
-                "--build-arg",
-                "TAG={}".format(mainTags[1]),
-            ]
-            if config.noEngine == False:
+            if config.buildTargetEngine:
```
slonopotamus, line 427: This condition is wrong. `ue4BuildArgs` are used below on line 453 to build `ue4-minimal`.

zorbathut: Good catch! Fixed, I think - let me know if it looks right.

slonopotamus (approved):

slonopotamus:
@adamrehn @TBBle Do you have any comments on this?

zorbathut:
Just FYI, I actually have another small pile of fixes and improvements incoming; want me to toss them all on here, or wait for this to go in and turn them into separate PR's?

(this covers your initial comment and some minor issues I ran into along the way)

TBBle:
Sorry, I'm currently snowed-under with other stuff, and haven't had a chance to give it a review. So no concerns from me at this time. ^_^

slonopotamus:
@zorbathut thanks for your efforts, I've merged this PR. You're very welcome to submit further improvements

PR #236, opened 2022-04-05
Add warning for `--no-*` target specifiers

zorbathut:
I am actually somewhat unhappy with this because it's not colored, but the `logger` object is created in `build.py` and does not yet exist at this point.

Better suggestions welcome.

### DIFF
--- a/ue4docker/infrastructure/BuildConfiguration.py
+++ b/ue4docker/infrastructure/BuildConfiguration.py
@@ -284,6 +284,9 @@ class BuildConfiguration(object):
             self.args.target = ["all"]
         elif using_target_specifier_old and not using_target_specifier_new:
             # Convert these to the new style
+            print(
+                "Warning: Using deprecated `--no-*` target specifiers; recommend changing to `--target`"
+            )
 
             # no-minimal implies no-full
             if self.args.no_minimal:


slonopotamus:
Hmm... We should move logger initialization to a more early/generic place. It is perfectly fine to log from non-build subcommands, for example.

zorbathut:
I'll give it a try. It'll take some time for me to get back to this, but it's on the TODO list.

TBBle:
Maybe consider moving to the standard `logging` module, with idiomatic
```
logger = logging.getLogger(__name__)
```
at module-scope (so you _always_ have a logger available), and to customise the output, configure the logging module from `main` with a formatter that looks like current `.infrastructure.Logger`'s output.

That would also give us a free `warning` level which we currently lack. It would also be nicer for any use-cases where people are importing ue4-docker as a module and using it that way.

That's not a small change though, although it could probably be _prototyped_ pretty easily inside `.infrastructure.Logger` since currently all our log messages (except the new one here) go through it anyway.

zorbathut:
I think I am practically not going to have time to work on a significant change like this this for a while; you're welcome to accept it as-is, reject it, or just let it sit around until someone gets to it. But please don't wait on me either way :)

slonopotamus:
See #245

PR #237, opened 2022-04-06
Add build target status output on startup

zorbathut:


zorbathut:
I am pretty sure the CI error is not actually my fault :)

PR #238, opened 2022-04-06
Do dependency analysis with a set

zorbathut:


slonopotamus:
Do you actually need a _separate_ set? You can just use keys of `self.buildTargets`. And you actually only store `True` as values of `self.buildTargets`, so effectively `self.buildTargets` already works as a `set`.

zorbathut:
`self.buildTargets` intentionally includes both True and False, so you can query it by value and still be protected from typos via KeyError exception (which I had one of while developing this code, in fact!) True means "being built", False means "not being built, but a thing that could conceptually be built".

For similar reasons I want to avoid putting "all" in it; "all" isn't a valid target, it's a metatarget.

I'm being paranoid about bugs and error detection, but I kinda always am, I've spent too much time in life debugging errors to take it casually.

In an ideal world I'd have some kind of more formal data-driven dependency setup and get rid of the hardcoded stuff, but there's just not enough images being built right now for that work to be worth doing.

PR #242, opened 2022-04-14
Avoid copying DLL files that already exist under Windows

### LINKED ISSUE #241: DLL files fail to copy under Windows Server 2022
In commit edcedca I added two new DLL files to the list that gets copied at the start of the `ue4-build-prerequisites` Dockerfile to fix a DDC generation crash when building Windows containers for Unreal Engine 5.0.0. That fix was tested under Windows Server 2019, but subsequent testing under Windows Server 2022 has uncovered a situation that we haven't previously encountered:

```
Step 6/12 : COPY --from=dlls     C:\Windows\System32\avicap32.dll     C:\Windows\System32\avifil32.dll     C:\Windows\System32\avrt.dll     C:\Windows\System32\d3d10warp.dll     C:\Windows\System32\D3DSCache.dll     C:\Windows\System32\dsound.dll     C:\Windows\System32\dxva2.dll     C:\Windows\System32\glu32.dll     C:\Windows\System32\ksuser.dll     C:\Windows\System32\mf.dll     C:\Windows\System32\mfcore.dll     C:\Windows\System32\mfplat.dll     C:\Windows\System32\mfplay.dll     C:\Windows\System32\mfreadwrite.dll     C:\Windows\System32\msacm32.dll     C:\Windows\System32\msdmo.dll     C:\Windows\System32\msvfw32.dll     C:\Windows\System32\opengl32.dll     C:\Windows\System32\ResampleDMO.dll     C:\Windows\System32\ResourcePolicyClient.dll     C:\Windows\System32\
failed to copy files: failed to copy file: remove \\?\Volume{f9811f13-fba7-48ea-bf2a-69b78bc77376}\Windows\System32\ksuser.dll: Access is denied.
[ue4-docker build] Error: failed to build image "adamrehn/ue4-build-prerequisites:ltsc2022-vs2019".
```

The issue is that the LTSC 2019 version of the Windows Server Core base image does not include `ksuser.dll` but the LTSC 2022 base image does, and the Docker `COPY` directive is refusing to overwrite files that already exist. (It's unclear whether this is intended behaviour for Windows containers or a bug on the part of Docker, since testing shows that the `COPY` directive will happily overwrite existing files in Linux containers.)

I have a fix implemented but I'd like some feedback on it, so I'll create a pull request for that.

adamrehn:
Fixes #241.

The initial fix I devised for this problem was to simply copy the DLL files into an intermediate directory and then use `xcopy /y` to copy them into System32, happily overwriting any existing files that the `COPY` directive refused to touch. This was the simplest and least intrusive implementation I could think of, but unfortunately it introduced an additional filesystem layer and increased the size of the `ue4-build-prerequisites` image by the aggregate filesizes of all of the DLL files. Given that the Windows container images are already massive and every megabyte counts, I decided that it was worth the effort of implementing a better solution that does not impact the final image size.

This is the higher-effort fix, whereby DLL files are copied into an extra build stage and duplicates are programmatically identified and removed prior to being copied into the final image. After writing a version of the deduplication script in Python I then remembered that the intermediate build stage lacks the interpreter, and rather than bloating the Dockerfile with additional directives to install Chocolatey and Python, I rewrote the deduplication script in PowerShell. I'm pretty happy with the result, but I'd appreciate feedback from @slonopotamus and/or @TBBle before merging this into master, in case either of you can think of a more elegant solution to this problem.

Alternative solutions that I considered included the introduction of `ARG` directives or Jinja template values to allow us to modify the list of DLL files depending on the selected base image tag, but that approach felt brittle and would potentially require ongoing maintenance if new Windows releases make further additions to the set of DLLs included in the Server Core base image. By contrast, the solution implemented in this PR addresses the problem in a generic way and does not introduce any additional maintenance burden.

### DIFF
--- a/ue4docker/dockerfiles/ue4-build-prerequisites/windows/Dockerfile
+++ b/ue4docker/dockerfiles/ue4-build-prerequisites/windows/Dockerfile
@@ -4,8 +4,13 @@ ARG DLLSRCIMAGE
 
 FROM ${DLLSRCIMAGE} as dlls
 
-FROM ${BASEIMAGE} as prerequisites
+FROM ${BASEIMAGE} as deduplication
 SHELL ["cmd", "/S", "/C"]
+
+{% if not disable_labels %}
+# Add a sentinel label so we can easily identify all derived images, including intermediate images
+LABEL com.adamrehn.ue4-docker.sentinel="1"
+{% endif %}
 
 # Gather the system DLLs that we need from the full Windows base image
 COPY --from=dlls `
@@ -29,7 +34,16 @@ COPY --from=dlls `
     C:\Windows\System32\opengl32.dll `
     C:\Windows\System32\ResampleDMO.dll `
     C:\Windows\System32\ResourcePolicyClient.dll `
-    C:\Windows\System32\
+    C:\GatheredDLLs\
+
+# Remove any DLL files that already exist in the target base image, to avoid permission errors when attempting to overwrite existing files with a COPY directive
+COPY remove-duplicate-dlls.ps1 C:\remove-duplicate-dlls.ps1
+RUN powershell -ExecutionPolicy Bypass -File C:\remove-duplicate-dlls.ps1
+
+# Copy the DLL files into a clean image
+FROM ${BASEIMAGE} as prerequisites
+SHELL ["cmd", "/S", "/C"]
+COPY --from=deduplication C:\GatheredDlls\ C:\Windows\System32\
 
 {% if not disable_labels %}
 # Add a sentinel label so we can easily identify all derived images, including intermediate images
--- a/ue4docker/dockerfiles/ue4-build-prerequisites/windows/remove-duplicate-dlls.ps1
+++ b/ue4docker/dockerfiles/ue4-build-prerequisites/windows/remove-duplicate-dlls.ps1
@@ -1,0 +1,11 @@
+$dlls = (Get-ChildItem "C:\GatheredDLLs\*.dll")
+foreach ($dll in $dlls)
+{
+	$filename = $dll.Name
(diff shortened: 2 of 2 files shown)


slonopotamus:
Assuming your reasoning is correct (`COPY` fails because target file exists), I cannot suggest any better solution that you made.

> allow us to modify the list of DLL files depending on the selected base image tag

Please, no)

slonopotamus:
> `COPY` fails because target file exists

Things are not *that* simple. You can easily copy `C:\License.txt`.

slonopotamus (approved):

adamrehn:
> Things are not *that* simple. You can easily copy `C:\License.txt`.

Interesting. I wonder if it's because the files being overwritten are from one of the base filesystem layers that are marked as foreign layers?

TBBle:
Foreign layer won't matter, since that information is no longer visible once the layers are extracted. It is possible that there's filesystem flags on the DLLs in System32, e.g., only TrustedInstaller is allowed to modify them. This wouldn't be container specific.

Another option is that, depending on how `COPY` is implemented, if it's actually starting the container, the file might be in-use?

(I haven't looked at the PR yet, I was idlying thinking about the (unexpected to me) COPY overwrite failure mentinoed in #241 and then saw the discussion was going on here)

slonopotamus:
> Another option is that, depending on how `COPY` is implemented, if it's actually starting the container, the file might be in-use?

Nope, `COPY` doesn't work that way. If you try to use Hyper-V isolation on a system where Hyper-V doesn't work (for example, because you're already inside a VM without nested virtualization), the thing that will fail is [`RUN`](https://github.com/adamrehn/ue4-docker/blob/v0.0.97/ue4docker/dockerfiles/ue4-build-prerequisites/windows/Dockerfile#L40). `COPY` before it happily passes.

PR #244, opened 2022-04-18
fix: don't try to delete .git if it's not present

jlsalmon:
When using `--opt source_mode=copy` with a custom `.dockerignore` that excludes the `.git` directory from the docker build context, the `rmdir .git` command fails with `The system cannot find the file specified.` since docker does not have that directory in context. This patch avoids this scenario with a simple existence check.


TBBle:
LGTM.

PR #252, opened 2022-05-11
Extend CI test matrix with older supported environments

slonopotamus:


setup.py
```diff
@@ -47,7 +50,10 @@
         "psutil",
         "requests",
         "semver>=2.7.9,<3.0.0",
-        "setuptools>=60.4.0",
+        # We want newer setuptools to avoid shim path bug on newer Pythons.
+        # See #231 and https://github.com/pypa/setuptools/issues/3001
+        # Unfortunately, that version of setuptools doesn't have Python 3.6 support
+        "setuptools>=60.4.0;python_version>='3.7.0'",
```
TBBle, line 56: Should there also be an unversioned `setuptools` requirement for Python 3.6? Or is that implicit in having `setup.py`?

slonopotamus: Good point. I'm not that familiar with `pip`/`setuptools` internals and not sure who exactly does dependency solving. Need to google it a bit, possibly you're right.

slonopotamus: After some googling, I come to the conclusion that you do not need to have unversioned dependency on `setuptools` inside `setup.py`.

There are more modern approaches (like `pyproject.toml`) that support pluggable build tools [and actually solve chicken-and-egg problem with `setup.py`], and _there_ you would need to specify that you want `setuptools` (and not `poetry`, for example).

jonpas:
Looking at the CI logs, it is only building with VS2017 at the moment. Maybe this is worth expanding to VS2019 and VS2022 (when I PR that) as another matrix level.

setup.py
TBBle: Makes sense. My question was prompted partly because all my current projects _do_ use pyproject.toml and Poetry, so I am used to having to explicitly specify the build-system backend.

PR #265, opened 2022-06-20
[Windows] Remove engine intermediate engine build products before committing ue4-minimal builder image

slonopotamus:
These products are not used after build step, but they take massive amounts of disk space: up to 350GBs for multi-platform (Win64+PS4+PS5+XB1+XSX) engine build

To make matters even worse, current Docker image commit algorithm on Windows additionally uses x2 of resulting image size on disk.

### DIFF
--- a/ue4docker/dockerfiles/ue4-minimal/windows/Dockerfile
+++ b/ue4docker/dockerfiles/ue4-minimal/windows/Dockerfile
@@ -34,7 +34,8 @@ RUN .\Engine\Build\BatchFiles\RunUAT.bat BuildGraph `
     -set:HostPlatformOnly=true `
     -set:WithDDC={% if excluded_components.ddc == true %}false{% else %}true{% endif %} `
     {{ buildgraph_args }} && `
-	(if exist C:\UnrealEngine\LocalBuilds\InstalledDDC rmdir /s /q C:\UnrealEngine\LocalBuilds\InstalledDDC)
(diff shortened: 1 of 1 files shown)


TBBle:
That's pretty nuclear since it is blowing away stuff that was already there from the ue4-source image, but I guess we don't care about that directory after that point, and deletions will produce a much smaller container image that additions.

It's a shame there's no way (AFAIK) to tell the builder "Ignore changes under this directory" when capturing a `RUN` result.

Does the Linux pipeline need/want the same improvement?

TBBle (approved):

slonopotamus:
> That's pretty nuclear since it is blowing away stuff that was already there from the ue4-source image

I don't have an easier way to wipe away build stuff. We already removed `.git`, so we can't `git clean -fdx`. Even if we could, there are GitDependencies-downloaded files that we can't easily separate from build products.

> Does the Linux pipeline need/want the same improvement?

On Linux, there are several points:

1. Appropriate Dockerfile `RUN` directive is not committed at all, thanks to BuildKit
2. There are [some COPY steps](https://github.com/adamrehn/ue4-docker/blob/v0.0.99/ue4docker/dockerfiles/ue4-minimal/linux/Dockerfile#L58-L67) that need to be adjusted somehow.

slonopotamus:
UPD: this change doesn't save size of VHDX itself (because VHDX doesn't shrink when you remove files), but it does save 2x of build products size during image commit. And obviously, image commit is *much* faster since it doesn't need to write all those GBs to disk.

PR #266, opened 2022-06-23
Add labels, so it is possible to identify layers with engine

slonopotamus:
Intended scenario:
Suppose there is a tool that downloads and extracts engine from Docker image.
Such tool would like to only download relevant layers.
It doesn't want to download build-prerequisites layer.
It might not want to download debug layer.

With labels, a tool can analyze image manifest and only pick those engine layers that it needs.

### DIFF
--- a/ue4docker/dockerfiles/ue4-minimal/windows/Dockerfile
+++ b/ue4docker/dockerfiles/ue4-minimal/windows/Dockerfile
@@ -55,16 +55,33 @@ FROM ${NAMESPACE}/ue4-build-prerequisites:${PREREQS_TAG}
 {% endif %}
 
 # Copy the Installed Build files from the builder image
+
+{% if not disable_labels %}
+LABEL com.adamrehn.ue4-docker.layer.binaries
+{% endif %}
 COPY --from=builder C:\UnrealEngine\LocalBuilds\Engine\Windows C:\UnrealEngine
+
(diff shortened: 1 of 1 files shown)


TBBle:
I've never seen this idiom before, using `LABEL` to describe individual layers. My initial reaction is that the `LABEL` should be _after_ the layer it describes, because "layer parent" is more-fixed than "layer child".

That said, does this actually _work_? Poking around in the docs, I understand that `LABEL` turns into _image manifest_ annotations (OCI) or .Config.Labels (Docker), rather than being stored per-layer as they are declared.

I see there's _optional_ `history` in the Docker and OCI Image JSON, are you planning on parsing those entries?

slonopotamus:
> My initial reaction is that the `LABEL` should be _after_ the layer it describes

I spend two hours thinking whether `LABEL` should be before or after, but failed to find arguments for any of the options, so this was chosen by a fair coin flip.

> I see there's _optional_ `history` in the Docker and OCI Image JSON, are you planning on parsing those entries?

I do understand that I am doing something unorthodox (and possibly illegal in 50% of US states) here, but it turns outs that YES, you can see this data through manifest! I possibly should post some PoC code that utilizes this `LABEL` stuff.

slonopotamus:
We ended up not using this thing, so I'm going to close this PR.

PR #274, opened 2022-10-31
Avoid downloading unneeded dependencies on Linux

slonopotamus:
Before this commit, `ue4-docker build 5.0.3 --target=minimal`:

* `ue4-source`: 67.2GB
* `ue4-minimal`: 54.5GB

After this commit, `ue4-docker build 5.0.3 --target=minimal`:

* `ue4-source`: 47.5GB
* `ue4-minimal`: 47GB


slonopotamus:
Ideally, I would love to still see [this code](https://github.com/EpicGames/UnrealEngine/blob/4.21.2-release/Engine/Source/Programs/GitDependencies/Program.cs#L174-L193) in UE. However, it was removed for unkown reason.

PR #276, opened 2022-11-15
feat: ability to specify docker org for baseimg

russkel:
At the moment nvidia is being slow at pushing out ubuntu 22.04 cudagl images and we've been forced to create our own. This patch lets users use their own organisation images instead of nvidia's.

This adds a `-baseorg` argument to the build command.

### DIFF
--- a/ue4docker/infrastructure/BuildConfiguration.py
+++ b/ue4docker/infrastructure/BuildConfiguration.py
@@ -14,9 +14,11 @@ DEFAULT_GIT_REPO = "https://github.com/EpicGames/UnrealEngine.git"
 
 # The base images for Linux containers
 LINUX_BASE_IMAGES = {
-    "opengl": "nvidia/opengl:1.0-glvnd-devel-{ubuntu}",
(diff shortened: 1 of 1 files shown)


slonopotamus:
I'm +1 on the spirit of this PR (make base image more configurable), bit -1 on actual implementation. It makes more sense to make base image _fully_ configirable. Many usecases do not involve rendering at all and would happily use `ubuntu-nn.mm` for exame, without any connection to nvidia stuff.

russkel:
I agree. I did this quick hack because I needed to quickly move on. Specifying the image in entirety is likely a better move.

russkel:
I will remake this to make it a baseimg override instead.

patrick-5546:
I think changing the base image to `nvidia:cuda` instead of `nvidia:cudagl` also works:
- https://github.com/adamrehn/ue4-docker/discussions/294#discussioncomment-4778592

russkel:
@patrick-5546 thanks a lot for bringing this to my attention. I have changed my build to use `cuda` now. You should create a PR for it ;).

PR #285, opened 2022-12-06
Stop installing .Net Core

slonopotamus:
UE4.x doesn't need it while UE5 bundles its own copy

----

UE-5.0.3 happily builds with this change. 5.0EA2 testing is in progress yet.


slonopotamus:
Weird. 5.0EA2 also builds. I'm not sure why I added `Microsoft.NetCore.Component.SDK` in the first place in 825f3f380c1167d68a6768e17e169d65f84a2ea5

TBBle:
Probably an artefact of Epic's CI process, and of the Installed Build being built in a self-contained and specifically-versioned CI environment, but targeted for use on developer boxes who're likely to have .NET installed already.

TBBle:
A bunch of that stuff is things that you can't actually bundle into the source, e.g. Visual Studio.

And there's probably stuff there where they don't care what version is used (or haven't cared _yet_, and once it starts biting them, they'll bundle it).

It could also be that there's no strategic direction there, and it depends on who last touched a dependency to determine whether the dependency gets added to the CI system image, or bundled into the source tree.

At my last employer, I saw all three of the above mechanisms in-play in one project, and that project was only a few years younger than UE.

PR #288, opened 2022-12-10
fix(package): patch filters to exclude hololens due to package error

CanisHelix:
Packaging in UE 5.1.0 (even outside of UE4-Docker) results in an error

```
Unhandled exception: One or more errors occurred. (Script module "C:\UnrealEngine\Engine\Platforms\Hololens\Binaries\DotNET\AutomationTool\AutomationScripts\Platforms\HoloLens\HoloLens.Automation.dll" not found for record "C:\UnrealEngine\Engine\Intermediate\ScriptModules\HoloLens.Automation.json")
```

The PR to fix this itself doesn't work, but patching `InstalledEngineFilters.xml` instead provide the same workaround.

### DIFF
--- a/ue4docker/dockerfiles/ue4-source/windows/patch-broken-releases.py
+++ b/ue4docker/dockerfiles/ue4-source/windows/patch-broken-releases.py
@@ -53,3 +53,30 @@ if (
             print("PATCHED {}:\n\n{}".format(gitdepsFile, gitdepsXml), file=sys.stderr)
         else:
             print("PATCHED {}".format(gitdepsFile), file=sys.stderr)
+
+
+# Determine if we are building UE 5.1.0
+if (
+    versionDetails["MajorVersion"] == 5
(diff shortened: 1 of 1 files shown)


slonopotamus:
See https://github.com/adamrehn/ue4-docker/blob/master/.github/CONTRIBUTING.md#autoformatting-your-changes for instructions on how to apply codestyle.

WRT PR itself - I'm afraid we also need to fix Linux build. I haven't tried building 5.1.x yet, but it is very likely it also suffers from this HoloLens issue.

CanisHelix:
Applied the styling despite that it make that specific file inconsistent between ' and " in different blocks.

Re: Linux though, it does not suffer the same problem. My project fails to package due to code issues on Linux, but proceeds well into the process. On Windows the process doesn't even get to compiling as the Hololens issues crops up almost immediately.

CanisHelix:
Hmmm it's failing Lint on a file I didn't even submit

PR #291, opened 2022-12-11
Add docs to the repo

slonopotamus:
The ultimate goal is to make it possible to keep docs in sync with code, by storing them in a single repo and updating both at once. Current state (with docs being hosted at https://gitlab.com/adamrehn/docs.adamrehn.com) leads to docs lagging behind because either nobody knows about that repo at all or forgets / too lazy to submit a separate PR there. As a result, we have docs that were not updated for a year by _anyone_.

This set of changes is far from finished.

1. Pages need to be updated to remove obsolete and add missing stuff. For example, remove EOL Windows versions, merge "advanced build options" into `ue4-docker build` page, get rid of pages that got moved to https://unrealcontainers.com/, sync commandline flags with what we actually have nowadays, stop suggesting Docker EE for Windows Server (there's no such thing), document 20GB COPY issue
2. It is very likely I've messed up markup somewhere
3. Man pages need to be generated and packaged to pip
4. Publishing in HTML form is missing
5. We possibly want to reorganize stuff
6. Redirects from https://docs.adamrehn.com/ue4-docker are needed
7. Etc

But we should start somewhere, right?)

----

There's _a bit_ of passive-aggressiveness in this PR: I'm changing markup language of the docs. But I believe Asciidoc is superior to Markdown from technical POV. There are no more hacks for section titles, for admonitions and tables. Also, we now can generate proper man pages. Asciidoc is successfully used by multiple big OSS projects for over a decade and has proved itself as a powerful instrument.


adamrehn:
Thoughts:

- After our [initial discussion](https://discordapp.com/channels/844118704941170688/844121719475535913/1035960832128983060) regarding docs migration, I thought about whether it's worth combining docs from multiple project repos into a single site as part of a CI/CD pipeline to retain the current structure and visual consistency, and ultimately decided that it's probably more hassle than it's worth. The top-level AdamRehn Docs site is already designed to act as a jumping-off point to a set of separate subsites, so I figure each subsite can just be a separate website going forward. (The motivating benefit of this approach is that each site can use completely different toolchains to generate their HTML, so the ue4-docker docs can migrate to whichever stack best meets the project's needs whilst other projects like ue4cli and conan-ue4cli can continue using their existing Jekyll-based stack for the time being.) I'm currently leaning towards using subdomains for each project's separate docs site, e.g. `ue4-docker.docs.adamrehn.com` in the case of ue4-docker.

- Regarding the transition from Markdown to Asciidoc: I'm always happy to use the best tool for the job, and looking through the Asciidoc website it definitely seems to have a number of advantages over Markdown. (I particularly like the ability to generate books in PDF form! I recently tried to implement support in my Jekyll docs theme for generating HTML that would print to PDF nicely through the use of CSS Paged Media, and it was a massive pain. I'm now thinking that I might abandon that effort and simply use Asciidoc instead for docs projects that want to provide a PDF download for offline use.) Would the intent be to use the [Jekyll AsciiDoc Plugin](https://github.com/asciidoctor/jekyll-asciidoc) to render out HTML using the existing Jekyll docs theme, or to migrate to using Antora for HTML generation?

- In terms of overhauling and updating the content, perhaps this initial PR could focus on migrating the existing content as-is, and then the overhaul/update could be implemented in subsequent commits/PRs? That could help break this massive undertaking down into more manageable pieces, and also make it easier for multiple people to assist with the changes to the content.

slonopotamus:
For the reference, this is an example of PDF that we can now produce: [index.pdf](https://github.com/adamrehn/ue4-docker/files/10205621/index.pdf)

slonopotamus:
Soo, we have

1. HTML version: https://slonopotamus.github.io/ue4-docker/
2. PDF version: https://slonopotamus.github.io/ue4-docker/ue4-docker.pdf
3. EPUB version: https://slonopotamus.github.io/ue4-docker/ue4-docker.epub

Note that `adamrehn/ue4-docker` repo requires some setup to make pages visible. Adam talked about custom domain (`ue4-docker.docs.adamrehn.com`), it looks like such thing is supported in GitHub Pages.

image

PR #300, opened 2023-03-05
Bundle code from `container-utils`

slonopotamus:
[container-utils](https://github.com/adamrehn/container-utils)

1. Isn't used by anyone except us
2. Isn't actively maintained
3. Forces `setuptools`/`wheel`/`twine` deps upon users
4. Overall, doesn't look like a valuable asset in itself

This commit brings code from `container-utils` into ue4-docker repo and drops dependency on `container-utils`.

### DIFF
--- a/src/ue4docker/test.py
+++ b/src/ue4docker/test.py
@@ -1,6 +1,16 @@
-from .infrastructure import DockerUtils, GlobalConfiguration, Logger
-from container_utils import ContainerUtils, ImageUtils
-import docker, os, platform, sys
+import ntpath
+import os
+import posixpath
+import sys
+
+import docker
+from docker.errors import ImageNotFound
+
+from .infrastructure import (
+    ContainerUtils,
+    GlobalConfiguration,
+    Logger,
+)
 
 
 def test():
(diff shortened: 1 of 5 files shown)


slonopotamus:
Actually, I've got a strong feeling that we already have other code that does almost everything these utils provide...

slonopotamus:
Done, this ended up to be much smaller.

src/ue4docker/test.py
```diff
@@ -14,45 +24,52 @@ def test():
...
         # Start a container to run our tests in, automatically stopping and removing the container when we finish
         logger.action(
-            'Starting a container using the "{}" image...'.format(image), False
+            'Starting a container using the "{}" image...'.format(image_name), False
+        )
+        container = ContainerUtils.start_for_exec(
+            client, image_name, platform, isolation=isolation
         )
-        container = ContainerUtils.start_for_exec(client, image, isolation=isolation)
         with ContainerUtils.automatically_stop(container):
```
TBBle, line 52: Probably for a future change, but idiomatically this should be one call, something like `with container = ContainuerUtils.exec_container(client, image_name, platform, isolation=isolation):`. One of the strengths of the context manager pattern is that it provides a way to guarantee a resource-cleanup opportunity if something inside `start_for_exec` throws an exception.

I also could be misremembering something we built locally in my last job, but I thought the docker client already had a context manager pattern for this kind of use-case.

adamrehn:
Looks good to me. For memory, some private projects were using the `container-utils` package a few years back, but only ever for hacky stuff that acted as a stopgap measure until a proper solution could be put in place. I should probably archive that repo (along with a bunch of other repos that haven't been touched in ages).

@slonopotamus lots of great refactoring happening lately (#297, #298, #299 and this), thanks for all your efforts here!

src/ue4docker/test.py
slonopotamus: Yep, I'm not happy with ergonomics of current API, but since we have only a single (and very rarely used) place where it appears, possibly it can live as it is for some time.

PR #320, opened 2023-05-11
Patch Commit.gitdeps.xml to use new CDN

slonopotamus:
See https://forums.unrealengine.com/t/upcoming-disruption-of-service-impacting-unreal-engine-users-on-github/1155880

In May 2023, Epics broke Commit.gitdeps.xml for *all existing releases up to (including) 5.1.1* due to changes in their CDN


slonopotamus:
This doesn't work because we need to authenticate in order to download `Commit.gitdeps.xml`.

PR #332, opened 2023-09-01
resolves #313 restore ue4-engine image

### LINKED ISSUE #313: Support Engine Images Again
### Origin:

Running `ue4-docker build --help` suggests `engine` as a valid target still after #273

### Possible Fixes:
- add the support for this target back (which I'm advocating for)
- remove `engine` from the help text

Before taking action, the two options should be weighed.

### Weighing the options

Although there may not seem any use in the engine image, there is a functionality difference between an installed build and a source build of unreal.

For example:
My team requires a source build so that unreal can package individual plugins. This feature breaks with an installed build. Packaging a plugin allows me to send downstream developers packaged versions of my plugin instead of the source code. I use CI to verify this process on every merge in all of our plugins.

If I loose the ability to use the engine image next time I update unreal versions, I will be forced to maintain dockerfiles and build process for these images, or an engine fork. Adding these tasks to my ever-increasing workload would squash any hope of timely upgrading engine versions and force my team (and downstream) to use ancient technology.

Stripping the help text would be significantly faster and easier, but that's why I'm opening an issue instead of a pull request.

I know it's a lot to ask, but it would mean a lot if support is continued for the engine image going forward.

If nothing else, I hope this issue challenges the assumption many have that installed and source builds have feature parity (apart from the ability to modify the engine source code).

In my case, the performance cost of an engine image doesn't matter. I don't need to build this image every day, nor am I short of data storage. Most use cases don't require an engine image because functional differences are rare and performance is often a more motivating constraint. Further, the developers who use this project the most frequently switch engine versions or rebuild these containers.

slonopotamus:
Also, drop deprecated for more than a year `--no-full`/`--no-minimal` options and just use `--target`.

### DIFF
--- a/docs/ue4-docker-build.adoc
+++ b/docs/ue4-docker-build.adoc
@@ -87,12 +87,6 @@ Monitor resource usage during builds (useful for debugging)
 *--no-cache*::
 Disable Docker build cache
 
-*--no-full*::
-Don't build the ue4-full image (deprecated, use *--target* _target_ instead)
-
-*--no-minimal*::
-Don't build the ue4-minimal image (deprecated, use *--target* _target_ instead)
-
 *--opt* _opt_::
 Set an advanced configuration option (can be specified multiple times to specify multiple options)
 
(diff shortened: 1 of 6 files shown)


slonopotamus:
I am not going to push this forward

PR #346, opened 2024-01-11
Follow Epic's logic for removing debug symbols

uhlenbrock:
There are `pdb` files, not prefaced with `UnrealEngine`, which are expected to be present when creating a shipping build.

*  C:\UnrealEngine\Engine\Binaries\Win64\tbb.pdb
*  C:\UnrealEngine\Engine\Binaries\Win64\tbbmalloc.pdb

/cc @slonopotamus

### DIFF
--- a/src/ue4docker/dockerfiles/ue4-minimal/windows/split-components.py
+++ b/src/ue4docker/dockerfiles/ue4-minimal/windows/split-components.py
@@ -48,7 +48,7 @@ ddc = [join(rootDir, "Engine", "DerivedDataCache", "Compressed.ddp")]
 extractComponent(rootDir, outputDir, "DDC", "Derived Data Cache (DDC)", ddc)
 
 # Extract debug symbols
-symbolFiles = glob.glob(join(rootDir, "**", "*.pdb"), recursive=True)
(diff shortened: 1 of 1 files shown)


adamrehn:
While performing testing for #359, I discovered that this change actually breaks 4.27 builds when not excluding debug symbols, hitting the following error:

```
 ---> c114ea905ab9
Step 29/34 : COPY --from=builder C:\UnrealEngine\Components\DebugSymbols C:\UnrealEngine
COPY failed: stat C:\UnrealEngine\Components\DebugSymbols: file does not exist
```

The new glob pattern doesn't match any PDB files at all under 4.27, so the `DebugSymbols` directory is empty. Evidently the `COPY` directive doesn't like merging an empty directory over an existing one, leading to the somewhat misleading error message shown above. Excluding debug symbols avoids this, but it doesn't actually reduce the image size, since all of the PDB files are left in their original locations and end up being copied into the final image stage. I've adjusted the glob pattern in commit 4e53548 to match both `UE4Editor` and `UnrealEditor`, and testing has confirmed that it works correctly for both Unreal Engine 4.27.2 and 5.3.2.

Issue #350, opened 2024-02-07
Proposal: limit supported Unreal Engine versions to N latest releases


adamrehn:
## Problem overview

The current approach to Unreal Engine version support in ue4-docker is that we try to encompass as many releases as is feasible from a technical standpoint, and we only drop support for old releases when we make changes that break compatibility. For example, we dropped support for Unreal Engine 4.20 when we migrated to using Ubuntu 22.04 base images by default as part of #296, and then more recently in commit 84ca65c I introduced code that requires Unreal Engine 4.22 or newer. (Even then, the breaking code could easily be gated behind a check if we wanted to maintain support for Unreal Engine 4.21.)

Supporting a broad range of past Unreal Engine releases is helpful for users who cannot migrate to a newer version (e.g. due to working on a large project that has pinned the version of UE that it is built upon), but it also hinders the development of ue4-docker because it significantly increases the friction involved when making potentially breaking changes. Any change that might break old releases needs to be tested against all supported releases, and the test matrix has grown to an unwieldy size. In commit f249d1e I added a [script to run image build tests for all supported releases](https://github.com/adamrehn/ue4-docker/commit/f249d1e8a49f5ba4096dc3778bf23e0f57120090#diff-f4dbae4f2cd12620bda61040161bd1e7ed927a2b897cb1d2176436796a4b9975), which is based on the testing I did as part of #296. Running this script in serial can take days, even on reasonably powerful hardware, and if the built images aren't automatically removed after each test then the disk space on most machines will be exhausted before completing all of the tests. Running the script in parallel reduces this time, but requires a sufficient number of machines or VMs, and that number only continues to grow as the list of supported releases grows. (Then multiply all of that by two if you're testing changes to both the Linux and Windows Dockerfiles.)

At the time of writing, ue4-docker supports the following Unreal Engine releases:

- 4.22, released in [April of 2019](https://www.unrealengine.com/en-US/blog/unreal-engine-4-22-released)
- 4.23, released in [September of 2019](https://www.unrealengine.com/en-US/blog/unreal-engine-4-23-released)
- 4.24, released in [December of 2019](https://www.unrealengine.com/en-US/blog/unreal-engine-4-24-released)
- 4.25, released in [May of 2020](https://www.unrealengine.com/en-US/blog/unreal-engine-4-25-released)
- 4.26, released in [December of 2020](https://www.unrealengine.com/en-US/blog/unreal-engine-4-26-released)
[...]

PR #370, opened 2024-12-11
Drop engines older than 4.27

slonopotamus:
See #350


README.md
```diff
@@ -9,7 +9,7 @@ The ue4-docker Python package contains a set of Dockerfiles and accompanying bui
 
 Key features include:
 
-- The six most recent versions of the Unreal Engine are supported (currently Unreal Engine 4.26 and newer).
+- The six most recent versions of the Unreal Engine are supported (currently Unreal Engine 4.27 and newer).
```
TBBle, line 12: This patch makes it 7 versions. Perhaps we just say "Six most recent plus 4.27-plus" (or whatever the official name for that 4.27-plus LTS branch is...) That's assuming we'll keep 4.27-plus support but drop 5.0 support when 5.6 is released.

src/ue4docker/dockerfiles/ue4-source/windows/patch-broken-releases.py
```diff
@@ -14,45 +14,12 @@ def writeFile(filename, data):
         f.write(data.encode("utf-8"))
 
 
-# Determine if we are building UE 4.25.4
+# Determine if we are building UE 4.27.0
```
TBBle, line 17: This comment no longer applies and can just be removed.

src/ue4docker/dockerfiles/ue4-minimal/linux/patch-build-graph.py
```diff
@@ -1,57 +0,0 @@
...
-    versionData["MajorVersion"] != 4
-    or versionData["MinorVersion"] != 24
-    or versionData["PatchVersion"] >= 3
-):
-    code = code.replace(
-        'Option Name="WithClient" Restrict="true|false" DefaultValue="false"',
-        'Option Name="WithClient" Restrict="true|false" DefaultValue="true"',
-    )
-    code = code.replace(
-        'Option Name="WithServer" Restrict="true|false" DefaultValue="false"',
```
slonopotamus, line 52: Possibly removal of this was too much aggressive...

slonopotamus: Moved [here](https://github.com/adamrehn/ue4-docker/pull/370/files#diff-629c9f3db0f071a8b003ae387eab4661453fdd465cf2fc68bdb4db8b3e970cd8R24). If user wants to disable Client/Server configurations, they can override our setting via `ue4-docker build --opt buildgraph_args=-set:WithClient=false -set:WithServer=false` as before this PR.

slonopotamus:
I believe this is ready to be merged. @TBBle wanna take a look?

PR #372, opened 2025-04-10
Using suggested MSVC build tools version for latest UE versions

Alexey31rus:
Because on the latest build tools, the compilation of the engine and/or project may fail due to errors. Therefore, we use the recommended version of the build tools for UE 5.4 and 5.5

### DIFF
--- a/src/ue4docker/dockerfiles/ue4-build-prerequisites/windows/install-prerequisites.ps1
+++ b/src/ue4docker/dockerfiles/ue4-build-prerequisites/windows/install-prerequisites.ps1
@@ -55,14 +55,36 @@ Expand-Archive -Path "$env:TEMP\vulkan-runtime-components.zip" -DestinationPath 
 Copy-Item -Path "*\x64\vulkan-1.dll" -Destination C:\Windows\System32\
 
 $visual_studio_build = $args[0]
+$unreal_engine_version = $args[1]
 
 if ($visual_studio_build -eq "15")
 {
     $windows_sdk_version = 18362
+    $windows_os_version = "Windows10SDK"
 }
 else
 {
     $windows_sdk_version = 20348
+    $windows_os_version = "Windows10SDK"
+
+    # NOTE: See suggested components https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Config/Windows/Windows_SDK.json
+    if ($unreal_engine_version.StartsWith("5.4"))
+    {
+        $windows_sdk_version = 22621
+        $vs_tools_version = "14.38.17.8"
+    }
+    elseif ($unreal_engine_version.StartsWith("5.5") -or $unreal_engine_version.StartsWith("5.6"))
+    {
+        $windows_sdk_version = 22621
+        $windows_os_version = "Windows11SDK"
+        $vs_tools_version = "14.38.17.8"
+    }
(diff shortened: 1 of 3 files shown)


slonopotamus:
We can use this as a quick hack. For a more long-term solution, I'd like to use `Windows_SDK.json` from Unreal.

See https://discord.com/channels/844118704941170688/844121719475535913/1310942558502912001 and previous messages in that thread.

src/ue4docker/dockerfiles/ue4-build-prerequisites/windows/install-prerequisites.ps1
```diff
@@ -55,14 +55,30 @@ Expand-Archive -Path "$env:TEMP\vulkan-runtime-components.zip" -DestinationPath
...
     $windows_sdk_version = 18362
+    $windows_os_version = "Windows10SDK"
 }
 else
 {
     $windows_sdk_version = 20348
+    $windows_os_version = "Windows10SDK"
+
+    # NOTE: See suggested components https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Config/Windows/Windows_SDK.json
+    if ($unreal_engine_version.StartsWith("5.4"))
```
slonopotamus, line 71: `$windows_os_version` is not set in this branch

src/ue4docker/dockerfiles/ue4-build-prerequisites/windows/install-prerequisites.ps1
```diff
@@ -85,15 +101,26 @@ $vs_args = @(
...
     "--add", "Microsoft.Net.ComponentGroup.DevelopmentPrerequisites",
     "--add", "Microsoft.NetCore.Component.SDK",
     "--add", "Microsoft.NetCore.Component.Runtime.3.1"
 )
 
+if (-not ([string]::IsNullOrEmpty($vs_tools_version))) {
+    $vs_args = $vs_args,
+    "--add", "Microsoft.VisualStudio.Component.VC.$vs_tools_version.x86.x64",
+    "--remove", "Microsoft.VisualStudio.Component.VC.Tools.x86.x64",
+    "--remove", "Microsoft.VisualStudio.Component.Windows10SDK.20348"
```
slonopotamus, line 116: Why we need `--remove`? There is no VS installation yet when our script is executed -> there is nothing to remove.

Alexey31rus: Hmm, yes, indeed this component does not appear by default anymore

slonopotamus:
Fails on CI because UE version is undefined when we're building `ue4-docker build --target=prerequisites`. I suggest putting this PR on hold while we're implementing `Windows_SDK.json`-based approach.

jonpas:
CMake searches for `VC/Auxiliary/Build/Microsoft.VCToolsVersion.default.txt` (where `VC` in our case is `C:/BuildTools/`).

Looks like installing a specific VC component only creates the `VC/Auxiliary/Build/Microsoft.VCToolsVersion.v143.default.txt` (mind `v143` in the file name).

Making a copy of that file to the expected file works. Maybe that is something we have to consider for @slonopotamus' PR as well.

_With this, `CMAKE_GENERATOR` does not need to be manually set either as CMake successfully finds Visual Studio on its own and sets it as default on Windows._

PR #386, opened 2025-09-30
patch linux older ue with CDN gitdeps.xml matching release

bradsaracik:
unmerged PR #320 by @slonopotamus failed with authentication. This I think gets past it, but I only have run it on 4.27.2 under Linux (and in files using ubuntu24 to come in a separate PR)
Apologies this is only in Linux. maintainers allowed to edit
Black ran on file.

Note, this may also remove the need for https://github.com/adamrehn/ue4-docker/blob/master/test-suite/test-ue-releases.py#L28

### DIFF
--- a/src/ue4docker/dockerfiles/ue4-source/linux/patch-broken-releases.py
+++ b/src/ue4docker/dockerfiles/ue4-source/linux/patch-broken-releases.py
@@ -2,6 +2,8 @@
 import json
 import sys
 from os.path import join
+import os
+import requests
 
 
 def readFile(filename):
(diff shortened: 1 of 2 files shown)


src/ue4docker/dockerfiles/ue4-source/linux/patch-broken-releases.py
```diff
@@ -42,3 +44,36 @@ def writeFile(filename, data):
         print("PATCHED {}:\n\n{}".format(buildFile, buildXml), file=sys.stderr)
     else:
         print("PATCHED {}".format(buildFile), file=sys.stderr)
+
+if major < 5 or (major == 5 and minor < 2):
+    gitdepsFile = join(engineRoot, "Engine", "Build", "Commit.gitdeps.xml")
+    # See https://forums.unrealengine.com/t/upcoming-disruption-of-service-impacting-unreal-engine-users-on-github/1155880
+    # In May 2023, Epics broke Commit.gitdeps.xml for *all existing releases up to 5.1.1* due to changes in their CDN
+    # we need to authenticate
+    password = os.getenv("GITPASS")
```
slonopotamus, line 53: But where does `GITPASS` come from?

